PR #22, opened 2023-08-30
Tiled algorithm implementations

graeme-a-stewart:
This is a big merge, which adds three implementations of the N2Tiled algorithm. These use different data layout strategies and have rather different performance.

The first uses an SoA layout connected to the tiles themselves, so that each tile's jets are stored in a compact structure. However, this requires many containers to be allocated (`Vector{Float64}` for each tracked parameter) across many tiles, which is really expensive in typical scenarios where there are O(500) tiled being tracked. This is the `N2TiledSoATile` strategy. It is the slowest of the 3 implementations.

The second uses a global SoA, for a compact jet representation. This is much faster for the global operations such as finding the minimum value for `dij`. However, keeping the compact global SoA requires shuffling and shrinking containers as jets are merged and finalised, which is quite a lot of data movement. This is the `N2TiledSoAGlobal` strategy. It is more than x2 faster than the SoA per tile, but still suffers from complex book keeping and data shuffling.

The third algorithm is a port of Philippe Gras' FastJet inspired implementation of the `N2Tiled` algorithm from that package. It keeps book keeping very light, but maintains a compact list of `dij` parameters to allow for fast searching for the jet merges and finalisations. This turns out to be optimal.

This PR is to snapshot the work done so far. It will then be the starting point for a production release of this package.


graeme-a-stewart:
Just accepting this one myself as we need this code in the repo for the CHEP paper

PR #31, opened 2023-10-25
Pre release cleanup

### LINKED ISSUE #5: use existing Lorentz Vector package
https://github.com/JuliaHEP/LorentzVectorHEP.jl instead of https://github.com/gojakuch/JetReconstruction.jl/blob/main/src/Particle.jl ?

### LINKED ISSUE #15: Optimisation needed
The current Julia implementation is significantly slower than C++ FastJet and I assume we can do better. Here you can see a plot that compares benchmarking results from an older version of this repo, which is already heavily optimised, ("Faster Julia" from about a week ago, showed in dark purple) with the currently used version of the algorithm ("New Faster Julia", in magenta).  C++ FastJet is still way below. I did not include compilation time and used BenchmarkTools.jl for benchmarking. For the Julia approaches I've plotted the entire range of runtimes as well as the median (it's very close to the lower bound).

image

I cannot think of any other methods to optimise the code for now. The latest updates have already been inspired mostly by @Moelf 's suggestions (i.e. using `@inbounds @simd for`, `Base.@propagate_inbounds`, etc). In case anyone is able to spot other issues or knows ways to optimise the code with more or less native Julia methods, please let me know.

P.S. Profiling shows that most of the time we spend in the two following blocks of code:
```julia
...
## src/Algo.jl @ line 105 
# initialize _nn
@simd for i in 1:N
    _upd_nn_crosscheck!(i, 1, i-1, _eta, _phi, _R2, _nndist, _nn)
end
...
```
```julia
...
## src/Algo.jl @ line 165 
# update nearest neighbours step
@inbounds @simd for k in 1:N
    _upd_nn_step!(i, j, k, N, Nn, _kt2, _eta, _phi, _R2, _nndist, _nn, _nndij)
end
...
```

### LINKED ISSUE #23: Define public interface
We should define what the public interfaces are to the package. At the moment we have:

## N2Plain

Defaults to a collection of arbitrary objects, where `px`, `py`, `pz`, `energy` are defined for each object in `JetReconstruction` to extract the relevant 4-vector parameters.

I like that, but the *default* used in `Particle.jl` is a little bit hokey for my taste as the index numbers are mapped to each quantity in a *magic number* kind of way.

## N2Tiled

A collection of `PseudoJet` structs.

The think a dislike currently is that the struct has an *implementation specific* parameter, viz. `_cluster_hist_index`.

## Proposal

Standardise to accept input particles as a collection of `LorentzVector`s, used via the `LorentzVectorHEP` package, which provides the appropriate `px`, `py`, `pz`, `energy` definitions already. This also achieves #5.

- I still like the idea of documenting for type `T` the idea of defining `px{T}`, etc. that provides a clear way for users to use their own types with the package.

### Internal Use

It is probably optimal that each algorithm uses an *internal EDM* when users pass arbitrary types, as these can be tuned for internal efficiency (e.g., the N2Tiled method can keep using its current `PseudoJet`).

### LINKED ISSUE #25: License Matters
We have a bunch of code that came from [AntiKt,jl](https://github.com/grasph/AntiKt.jl) that is licensed under GPL. The rest of the package is under the more permissive MIT license (more popular in the Julia community IMO).

Current CERN and HSF advice is to *not* use viral licenses and, specifically, ATLAS and CMS disfavour them (they use Apache2).

@grasph - would you be willing to relicense the code under an MIT license?

graeme-a-stewart:
This is a significant clean-up, taking us close to an actual release of this package.

The two underperforming algorithms (combining tiled and SoA) have been removed.

The two remaining algorithms have had their main APIs brought into line with one another:
- named as `plain_jet_reconstruct(particles::Vector{T}; p = -1, R = 1.0, recombine = +, ptmin = 0.0)` and `tiled_jet_reconstruct(particles::Vector{T}; p = -1, R = 1.0, recombine = +, ptmin = 0.0)`.

For the input particle type, `JetReconstruction.PseudoJet` and `LorentzVectorHEP` types are supported, but, in general, any type can be used that supports appropriate 4-vector methods (`pt2()`, `phi()`, `rapidity()`, `px()`, `py()`, `pz()`, `energy()`).

The return type of both algorithms is `JetReconstruction.PseudoJet` as this is the most complete and efficient type - translation to `LorentzVectorHEP` is trivial, but always doing it is rather expensive (particularly as cylindrical vectors from `LorentzVectorHEP` store *pseudorapidity* and jet reconstruction uses *rapidity*).

The old style `Particle.jl` type is removed.

Significant amounts of renaming of files, methods and variables have been done to help clarity.

The `README` is reworked to be more useful.

There are still some defects that need to be addressed (see open issues), but a lot is now fixed.

Closes #5 (LorentzVectorHEP is supported, but is problematic to use internally)
Closes #15 (I think we got really pretty close to optimal now)
Closes #23 (API is well defined now, as is use of alternative input types)
Closes #25 (MIT everywhere)


src/Pseudojet.jl
```diff
@@ -138,13 +115,42 @@ pt(p::PseudoJet) = sqrt(p._pt2)
...
     x = m2(p)
     x < 0. ? -sqrt(-x) : sqrt(x)
 end
 
+# Ensure we have accessors for jet parameters
+px(p::PseudoJet) = p.px
+py(p::PseudoJet) = p.py
+pz(p::PseudoJet) = p.pz
+mass(p::PseudoJet) = m(p)
+mass2 = m2
```
Moelf, line 151:
```suggestion
const mass2 = m2
```

Moelf:
I'm not familiar with the content and FastJet APIs so can't say much :P

graeme-a-stewart:
> I'm not familiar with the content and FastJet APIs so can't say much :P

Thanks for spotting the non const issue!

> 
> But seems to be a huge step towards being a proper library package in terms of organization

Thank you! I hope no obvious defects or daft things done.

Moelf (approved):

Issue #26, opened 2023-09-22
Improve tests


graeme-a-stewart:
There is a nice benchmark of 100 Pythia 8 events that @grasph generated that we use for correctness and benchmarking ATM. 

I would propose that we extend this to two additional use cases:

- low density FCCee type events (where N2Plain might outperform N2Tiled)
- very high density HI type events (which will just be challenging)

In both cases FastJet should be used as a reference to ensure correctness.

We should also test other clustering strategies, at least Cambridge/Achen ($p=0$) and Inclusive $k_T$ ($p=1$).

graeme-a-stewart:
This was fixed in 47182d9ac470c81c549c3eb20e4b7c3fd99ce6ed. All 3 main algorithms are tested, as well as different input types: `PseudoJet` and `LorentzVector`.

PR #32, opened 2023-10-30
Enable multiple algorithms to work with N2Tiled strategy

graeme-a-stewart:
Fix the tiled strategy implementation to respect the value of `p` which is passed, allowing switching between different algorithms (Anti-kt, Cambridge/Achen, Inclusive-kt).

Tests have been reworked to test both strategies for these three algorithms against the outputs from FastJet (which partly addresses #26).


graeme-a-stewart:
All looks good, so self-merging

PR #35, opened 2024-04-03
Generic strategy switching

### LINKED ISSUE #27: Nicer strategy switching
ATM, the algorithm strategy is manual - `--strategy=N2Plain` in `chep.jl`. The string is mapped to an `enum` in Julia, but there is an annoying manual mapping from the option string to the enum. I could not find a way to extract a nice string from the enum names, did I miss something?

However, reading about this issue, it seems people often use types in these cases, not enums. Should see if that is a nicer way. If a concrete type was used to map to each algorithm it could contain a user friendly string for the CLI.

### LINKED ISSUE #28: Defaut strategy
We should have a default strategy (like `Best`) where the number of input particles is used to pick the more optimal algorithm (FastJet does this). #26 would help us to empirically determine the breakpoints as a function of the density of initial particles.

graeme-a-stewart:
Improve the use of Emums that provide the different strategy switches. We use `EnumX` to have these scoped, otherwise generic names like `Best` would be troublesome. CLI string values are automatically mapped to the different strategy names.

A `Best` strategy is implemented that switches between `N2Tiled` and `N2Plain` based on empirical measurements.

The demonstration code is simplified by always calling the `generic_jet_reconstruct` function, delegating the strategy decision to inside the package (where it should be!).

Closes #28.
Closes #27.


graeme-a-stewart:
Self merging for further development

Issue #43, opened 2024-04-19
Plotting support


graeme-a-stewart:
Based on the new `ClusterSequence` return type, develop plotting routines to visualise the output of the reconstruction

gojakuch:
hi, so is this issue on the branch main or on the cluster-sequence?

graeme-a-stewart:
Hi @gojakuch. The `cluster-sequence` branch was rebased and merged, so it shouldn't have been hanging around - I have deleted it now.

My current development branch is `inclusive-jets`, which I have pushed to GitHb now, so you can see it. There are some changes in the pipeline to [`ClusterSequence.jl`](https://github.com/JuliaHEP/JetReconstruction.jl/blob/exclusive-jets/src/ClusterSequence.jl), adding algorithm and strategy information - plus the algorithm and strategy are encoded now as named enums (with the `EmumX` package). However, I don't think that changes anything w.r.t. the core visualisation task, which would use `ClusterSequence.jets` and `ClusterSequence.history`.

Summary: it should be ok to attack this on `main`!

My hope is to merge the `inclusive-jets` branch soon - it's in mostly working, but there are some differences with FastJet for the $d_{ij}$ cuts that I need to understand before it's really good to go (the $n_{jets}$ cut seems to work just fine). That said, next week I am at a workshop all week, so time is pretty limited for the rest of May 😅.

graeme-a-stewart:
Thanks so much @gojakuch and to @Moelf for helping! (I was completely overwhelmed last week at the HSF workshop)

PR #46, opened 2024-05-09
Fix plotting for `ClusterSequence` and update serialisation

gojakuch:
This PR tackles the issue #43 by implementing an additional variant of the `jetsplot` function that works with `ClusterSequence` on top of the existing old one that has been developed in 2022 at the beginning of the project. Additionally, I have updated the serialisation methods, as they were partially deprecated and `savejets` was incompatible with the new things in the repository.

### DIFF
--- a/src/Serialize.jl
+++ b/src/Serialize.jl
@@ -35,7 +35,7 @@ function savejets(filename, jets; format="px py pz E")
 end
 
 """
-    loadjets!(filename, jets; splitby=isspace, constructor=(x,y,z,E)->LorentzVector(E,px,py,pz), dtype=Float64) -> jets
+    loadjets!(filename, jets; splitby=isspace, constructor=(px,py,pz,E)->LorentzVector(E,px,py,pz), dtype=Float64)
 
 Loads the `jets` from a file. Ignores lines that start with `'#'`. Each line gets processed in the following way: the line is split using `split(line, splitby)` or simply `split(line)` by default. Every value in this line is then converted to the `dtype` (which is `Float64` by default). These values are then used as arguments for the `constructor` function which should produce individual jets. By default, the `constructor` constructs Lorentz vectors.
 
@@ -47,7 +47,7 @@ loadjets!("myjets1.dat", jets)
 loadjets!("myjets2.dat", jets)
 ```
 """
-function loadjets!(filename, jets; splitby=isspace, constructor=(x,y,z,E)->LorentzVector(E,px,py,pz), dtype=Float64)
+function loadjets!(filename, jets; splitby=isspace, constructor=(px,py,pz,E)->LorentzVector(E,px,py,pz), dtype=Float64)
     open(filename, "r") do file
         for line in eachline(file)
             if line[1] != '#'
--- a/src/Serialize.jl
+++ b/src/Serialize.jl
@@ -1,19 +1,17 @@
 """
 `savejets(filename, jets; format="px py pz E")`
 
-Saves the given `jets` into a file with the given `filename`. Each line contains information about a single jet and is formatted according to the `format` string which defaults to `"px py pz E"` but can also contain other values in any order: `"pt2"` or `"kt"` for transverse momentum, `"phi"` for azimuth, `"eta"` or `"rapidity"` for pseudorapidity, `"m"` for mass. It is strongly NOT recommended to put something other than values and (possibly custom) separators in the `format` string.
+Saves the given `jets` into a file with the given `filename`. Each line contains information about a single jet and is formatted according to the `format` string which defaults to `"px py pz E"` but can also contain other values in any order: `"pt2"` for pt^2, `"phi"` for azimuth, `"rapidity"` for rapidity. It is strongly NOT recommended to put something other than values and (possibly custom) separators in the `format` string.
 """
 function savejets(filename, jets; format="px py pz E")
     symbols = Dict(
         "E" => JetReconstruction.energy,
+        "energy" => JetReconstruction.energy,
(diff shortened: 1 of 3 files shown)


src/Serialize.jl
```diff
@@ -1,19 +1,20 @@
...
         "px" => JetReconstruction.px,
-        "pt" => JetReconstruction.pt,
-        "kt" => JetReconstruction.pt,
+        "pt2" => JetReconstruction.pt2,
+        "kt" => JetReconstruction.pt2,
         "py" => JetReconstruction.py,
         "pz" => JetReconstruction.pz,
         "phi" => JetReconstruction.phi,
         "m" => JetReconstruction.mass,
-        "eta" => JetReconstruction.eta
```
Moelf, line 16: should eta be eta?

src/Serialize.jl
```diff
@@ -1,19 +1,20 @@
...
+Saves the given `jets` into a file with the given `filename`. Each line contains information about a single jet and is formatted according to the `format` string which defaults to `"px py pz E"` but can also contain other values in any order: `"pt2"` or `"kt"` for transverse momentum, `"phi"` for azimuth, `"eta"` or `"rapidity"` for pseudorapidity, `"m"` for mass. It is strongly NOT recommended to put something other than values and (possibly custom) separators in the `format` string.
 """
 function savejets(filename, jets; format="px py pz E")
     symbols = Dict(
         "E" => JetReconstruction.energy,
         "px" => JetReconstruction.px,
-        "pt" => JetReconstruction.pt,
-        "kt" => JetReconstruction.pt,
```
Moelf, line 11: should `"kt" => pt`?

src/Serialize.jl
```diff
@@ -1,19 +1,20 @@
...
+        "pt2" => JetReconstruction.pt2,
+        "kt" => JetReconstruction.pt2,
         "py" => JetReconstruction.py,
         "pz" => JetReconstruction.pz,
         "phi" => JetReconstruction.phi,
         "m" => JetReconstruction.mass,
-        "eta" => JetReconstruction.eta
+        "eta" => JetReconstruction.rapidity,
```
Moelf, line 16: should eta be eta?

src/Serialize.jl
```diff
@@ -1,19 +1,20 @@
...
+Saves the given `jets` into a file with the given `filename`. Each line contains information about a single jet and is formatted according to the `format` string which defaults to `"px py pz E"` but can also contain other values in any order: `"pt2"` or `"kt"` for transverse momentum, `"phi"` for azimuth, `"eta"` or `"rapidity"` for pseudorapidity, `"m"` for mass. It is strongly NOT recommended to put something other than values and (possibly custom) separators in the `format` string.
 """
 function savejets(filename, jets; format="px py pz E")
     symbols = Dict(
         "E" => JetReconstruction.energy,
         "px" => JetReconstruction.px,
-        "pt" => JetReconstruction.pt,
-        "kt" => JetReconstruction.pt,
+        "pt2" => JetReconstruction.pt2,
+        "kt" => JetReconstruction.pt2,
```
Moelf, line 11: should "kt" => pt?

src/Serialize.jl
gojakuch: well when I first wrote that my idea was that we demand from the user to define eta() for their type. right now, according to the README.md, only pt2(), phi(), rapidity(), px(), py(), pz(), energy()  are required

gojakuch: and as far as I remember, eta is rapidity (I could be wrong here though)

src/Serialize.jl
gojakuch: pt() is no more required to be defined and there's now only pt2. I understand the confusion though

gojakuch: if it's absurd from the standpoint of physics then we could remove kt from the map entirely

src/Serialize.jl
Moelf: eta is pseudorapidity, the two are only strictly equal when the object has `mass = 0`

src/Serialize.jl
Moelf: I thought "kt" is transverse momentum, thus should equal to pt not pt2 (which is sqaure of that?)

src/Serialize.jl
gojakuch: doesn't the rapidity actually give the same thing in the code? I honestly don't remember these details already

src/Serialize.jl
gojakuch: well, as I've said I forgot how this works because I haven't been doing this physics in a while :)
I think addressing both of your comments, I could just leave it as it was, since it won't really crash unless someone tries to use "kt" in formatting without defining the kt() method for their type

gojakuch: nevermind, it actually will crash. I'll just stick to the set of required methods then: pt2(), phi(), rapidity(), px(), py(), pz(), energy()

gojakuch:
should be alright now

Moelf:
LGTM!

Moelf:
Do we wanna do like a "round trip" as unit test? Or that doesn't make much sense

gojakuch:
> Do we wanna do like a "round trip" as unit test? Or that doesn't make much sense

what do you mean? that we put data loading -> algo execution -> plotting -> result saving in a unit test?

Moelf:
I just mean that `x == loadjets(savejets(x))`?

gojakuch:
oh, that could make sense

gojakuch:
I think we're good to go with this one

Moelf (approved):

PR #48, opened 2024-06-04
Improved examples and dependencies

### LINKED ISSUE #42: Simpler examples
The current example is a mish-mash of code profiling and performance measurements. That should stay, but a much simpler example should be given.

A few notebook examples would also be very nice to have.

graeme-a-stewart:
Update examples to provide better stand alone demonstrations of the use of the package.

In particular have a very simple `jetreco..jl` example that shows the basic use of the package. All of the supplementary code for timing and profiling is in the `instrumented-jetreco.jl` example. These scripts are steered now by algorithm name, not my power value for the distance metric.

Also add a script and notebook example for jet visualisation.

Fix the package dependencies to be only the *real* package dependencies, with compatibility versions given. The examples now have their own separate `Project.toml` file with their heavier requirements.

Move the visualisation code to a package extension, which is loaded when `Makie` is found in the environment. This bumps the minimum Julia version to 1.9, which is needed to support this feature.

Update documentation for these changes.

Closes #42


graeme-a-stewart:
I think this is pretty solid (tests pass!), but if you have a chance to take a look I'd be grateful @hegner @gojakuch.

N.B. particularly for @gojakuch the visualisation code is now a model extension, meaning it is only loaded if `Makie` is present.

PR #56, opened 2024-06-07
Code formatting

graeme-a-stewart:
Use code formatting with the setup:

```toml
style = "sciml"
yas_style_nesting = true
```

Source files are reformatted and a GitHub action is added

### DIFF
--- a/.JuliaFormatter.toml
+++ b/.JuliaFormatter.toml
@@ -1,0 +1,2 @@
+style = "sciml"
+yas_style_nesting = true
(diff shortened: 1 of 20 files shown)


Moelf (approved):

Moelf:
This is the expected behavior or the format CI right?

graeme-a-stewart:
> This is the expected behavior of the format CI right?

Yes, when there is a file that's badly formatted - though the error message isn't super helpful, right?

Anyway, fixed the last file and let's retry...

Moelf:
I thought https://github.com/julia-actions/julia-format would give you inline comment or something

graeme-a-stewart:
It seems the workflow suggestion is to pipe this into https://github.com/reviewdog/reviewdog, which will then make a suggestion on the PR

Issue #68, opened 2024-07-16
Support retrieval of jet constituents


graeme-a-stewart:
For substructure studies, it's necessary to go from a final jet back to its constituent clusters.

For this, we need to (internally) map from the jet's index, through the history vector, to find all of the original cluster particle indexes.

Then we should return these particles to the user.

One question is, however, how the user will find their selected "final" jets, and retain enough information to find the mapping into the cluster sequence to perform this operation. The history index of the jet index are essentially internal bookkeeping objects that shouldn't really be exposed.

graeme-a-stewart:
Note on how FastJet does this - it works with Pseudo jet objects that keep a link to `_cluster_hist_index`, which is then their entry point. 

Note that our current functions to retrieve `exclusive_jets` and `inclusive_jets` return `LorentzVectorCyl`, so more information is required in this case.

Issue #49, opened 2024-06-04
Add support for Generalised k_T in reconstruction interfaces


graeme-a-stewart:
Currently the main interface to the two reconstruction strategies takes the power for the $d_{ij}$ metric as an argument, which then switches between the anti-kt, CA and kt versions (-1, 0, 1).

However, this is only good for these $pp$ algorithms and doesn't account for, e.g., the $e^+e^-$ algorithms, such as the generalised $k_T$.

So it's better to switch to steering with the explicit algorithm name and not a power value as a proxy.

grasph:
Hello Graeme. I don't understand your point. By definition in the generalized kt algorithm the exponent is a parameter. 

Named switch can be added for convenience, but to support the generalized algorithm, we need the exponent to be a  parameter (p ∈ℝ) when the switch is set to generalized_kt.

For e+e-, we need an extra switch to use E instead of pt, and change (ΔRij/R)² term by the one used for e+e-.

graeme-a-stewart:
Hi @grasph - thanks for he comment. Mathematically I agree with you, but does anyone ever use an arbitrary power? I only ever saw -1, 0, 1. Of course, we can continue to support arbitrary powers if they are desired.

Indeed, for the e+e- algorithms (#50 #51) I think a lot of code can be shared with the current pp algorithms, if we make the appropriate measures parameters.

grasph:
I'm not aware of use of different power parameter values than -1, 0, 1, but as supporting any parameter value comes from free, I think we should not restrict the software to these values.

This generalization that formulates in a consistent manner the different historical algorithms is very elegant. 

I suggest defining constants kt_power_parameter, antikt_power_parameter and cambridge_aachen_power_parameter, to be used as the power parameter, instead of adding a parameter selecting the algorithm and ending with two ways  to select the same algorithm (specific algorithm / generalized_kt with proper power parameter).

graeme-a-stewart:
Updating this issue to support properly a generalised k_T algorithm for pp in `(y, \phi)`.

- Add new reconstruction algorithm type, `GenKt`
- For this algorithm this type, plus a specific `p` value needs to be passed

For the fixed power algorithms, AntiKt, CA and Kt these should be specifiable either via their power value or their algorithm type (if both are given these must be consistent).

PR #70, opened 2024-07-18
Jet constituents implementation

graeme-a-stewart:
This PR addresses #68 in particular

**API Change**: `exclusive_jets` is changed so that `ptmin` is a named parameter
- This is more consistent with `inclusive_jets` and is better when the new optional type parameter is also given

Add `constituents()` function that allows retrieval of the original clusters of a jet (as `PseudoJets`)

The `exclusive_jets` and `inclusive_jets` can now specialise their return types, allowing them to take a parameter that defines the return type (in particular `PseudoJet`).
Note that only `LorentzVectorCyl` and `PseudoJet` are supported.

Add two examples showing how to use this.

Related to #68 (but not yet complete)


graeme-a-stewart:
Merging this now, even though full development is not yet complete, because other developments, e.g., to support Generalised kT (#49) are touching the same files and will result in merge conflicts.

`constituents()` should not be considered a stable interface yet.

PR #71, opened 2024-07-29
Generalised kT Algorithm for pp collisions

### LINKED ISSUE #49: Add support for Generalised k_T in reconstruction interfaces

graeme-a-stewart:
Add proper support for the generalised kT algorithm for pp collisions (i.e., operating in $(y, \phi)$ space).

Added a `JetAlgorithm` type for generalised kT, viz. `GenKt`.

Strategies are restructured to be able to be driven by only the power, or the algorithm, or (for `GenKt`) both. A new function, `get_algorithm_power_consistency` is added that will return the "partner" of a (power/algorithm) given an (algorithm/power). If inconsistent input is given an exception is thrown.

Documentation is updated explaining how to call with either `p` or `algorithm`, or both for `GenKt`.

ClusterSequence now stores the power value and the algorithm and has to be initialised with both. In addition the power is always stored as a `Float64`.

Examples are restructured to also be able to be driven by power, or algorithm, or both. However, if neither is given then the scripts will use `AntiKt` and print a warning. There is also now a clearer separation between calling the jet reconstruction itself and getting the inclusive/exclusive constituents.

A test to compare Generalised kT output with FastJet for the case `p=1.5` is added.

Tests are also added to run our code examples, to check none of these break (at least for the standalone Julia scripts).

# Misc

- Notebooks are now *not* formatted by JuliaFormatter.
- For the tiled algorithm directly calling the real work interface is now discouraged (method is now `_tiled_jet_reconstruct`) to ensure that power and algorithm are handled properly in the public interface (`tiled_jet_reconstruct`).

Closes #49


graeme-a-stewart:
There were non-trivial issues in running the examples tests, so these are put onto the backburner for now:

- The dependencies are heavy and some for GLMakie actually seem to timeout
- As these are CI machines, they can't correctly setup GLMakie
- There seems to be a weird issue with setting up a different environment for the Julia nightly:


```
ERROR: LoadError: ArgumentError: Package Pkg does not have Random in its dependencies:
- Note that the following manifests in the load path were resolved with a potentially
  different DEV version of the current version, which may be the cause of the error.
  Try to re-resolve them in the current version, or consider deleting them if that fails:
    /tmp/jl_qr3gn6/Manifest.toml
- You may have a partially installed environment. Try `Pkg.instantiate()`
  to ensure all packages in the environment are installed.
- Or, if you have Pkg checked out for development and have
  added Random as a dependency but haven't updated your primary
  environment's manifest file, try `Pkg.resolve()`.
- Otherwise you may need to report an issue with Pkg
```

I currently have no insight into that one.

The better plan is to restructure the examples to be able to run the subset which are CI friendly, thus move away from a monolithic `Project.toml` for all tests.

PR #80, opened 2024-10-08
Add support for EDM4hep inputs

graeme-a-stewart:
Add an extension that allows `ReconstructedParticles` to be read from EDM4hep files.

Ensure that internal jet objects can be initialised from any `Real`, not just `Float64` (which is too restrictive!).

For input particle collections, type to `AbstractArray{T, 1}` instead of `Vector{T}`, so that alternative containers can be used (specifically `StructArrays.StructVector` from EDM4hep).

Add documentation updates for this change. Unfortunately also retreat from Vitepress as it seems to completely screw up the ability to preview changes locally (locally generated pages are just broken).


graeme-a-stewart:
Going to merge this PR now, but TODO is to add a test with a small EDM4hep ROOT file (and generally work on code coverage in the tests).

PR #81, opened 2024-10-17
Fix significant performance regression

graeme-a-stewart:
When generalising `ClusterSequence` to also handle the `EEJet` type, the `jets` member was changed to be the *abstract* type `FourMomentum`.  This is type unstable and has a very negative impact on performance!

Fix `ClusterSequence` to be *parameterised*, where the type `T` is taken from the type of the jets passed in to the constructor.

This improves performance for $pp$ and $e^+e^-$ significantly.

Further fixes are to factorise out some of the `StructArray` handling functions in `EEAlgorithm.jl` to make the code logic easier to follow.


Moelf (approved):

graeme-a-stewart:
Thanks for checking @Moelf 👍

PR #82, opened 2024-10-23
Update Documentation.yml

### LINKED ISSUE #78: Fix documentation versions
At the moment we don't have versioned documentation correctly produced - we only get the `main` branch, i.e., `dev` documentation.

@Moelf pointed to [Lux.jl](https://lux.csail.mit.edu/stable/) as an example that has managed to get this correct - it's probably a minor configuration issue.

However, documentation for older releases may need to be created by hand.

Moelf:
fix #78

I think the problem is still that we don't have `DOCUMENTER_KEY` set in the repo Settings

### DIFF
--- a/.github/workflows/Documentation.yml
+++ b/.github/workflows/Documentation.yml
@@ -20,10 +20,8 @@ jobs:
         with:
           version: '1'
       - uses: julia-actions/cache@v2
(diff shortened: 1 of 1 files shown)


graeme-a-stewart:
Thank you @Moelf! 🤩

PR #87, opened 2024-11-05
Implementation of substructure modules

sattwamo:
Added functions for mass drop, soft drop, jet filtering and jet trimming in `src/Substructure.jl`
Added examples of using mass drop and jet trimming in `examples/jetreco-substructure.jl`


Moelf:
>The format action returns true when the code conforms (and it fixes is when it isn't). Our preferred format is in .JuliaFormatter.toml.

what's the status of making this inline comments of PR

graeme-a-stewart:
> > The format action returns true when the code conforms (and it fixes is when it isn't). Our preferred format is in .JuliaFormatter.toml.
> 
> what's the status of making this inline comments of PR

On the TODO list!

src/JetReconstruction.jl
```diff
@@ -89,4 +89,8 @@ export jetsplot, animatereco
 include("JSONresults.jl")
 export FinalJet, FinalJets
 
+# Substructure modules
+include("Substructure.jl")
+export MassDropTagger, SoftDropTagger, Filter, Trim, massDrop, softDrop, jetFilter, jetTrim
```
graeme-a-stewart, line 94: `Filter` and `Trim` are way too generic names to export into the global namespace. They should be called `JetTrim` and `JetFilter` IMO.

We should stick to the standard Julia naming schemes for methods, lower case and with `_` if needed. So `massDrop` -> `massdrop`, etc.

src/Substructure.jl
```diff
@@ -0,0 +1,313 @@
...
+Checks if the jet `p` is a child of two other jets, after clustering 
+
+# Arguments
+- `p`: The jet to check.
+- `history`: The vector of history element obtained from the cluster sequence after clustering.
+
+# Returns
+- (boolean, Int64, Int64): true or false depending on if the jet has a parent or not. If the jet has a parent, returns the indices of the parent jets in the history element. Otherwise, returns -2 (NonexistentParent).
+"""
+function has_parents(p::PseudoJet, history::Vector{HistoryElement})
```
graeme-a-stewart, line 13: This signature is ok as long as the function is *internal*. Anything the user would call should used a `ClusterSequence` object, so that we can evolve the implementation if we want to.

src/Substructure.jl
```diff
@@ -0,0 +1,313 @@
...
+Function to calculate the distance in the y-ϕ plane between two jets `jet1` and `jet2`
+
+# Arguments
+- `jet1`: The first jet.
+- `jet2`: The second jet.
+
+# Returns
+- Float64: The Euclidean distance in the y-ϕ plane for the two jets.
+"""
+function delta_R(jet1::PseudoJet, jet2::PseudoJet)
```
graeme-a-stewart, line 36: The jet types should not be restricted to `PseudoJet`, they should be `T` where `T <: FourMomentum`,

(This will evolve a bit later, but for now this is correct)

graeme-a-stewart, line 36: This function can be replaced by `sqrt(_tj_dist(jet1, jet2)` from `TiledAlgoLL.jl` (L27)

src/JetReconstruction.jl
```diff
@@ -89,4 +89,8 @@ export jetsplot, animatereco
 include("JSONresults.jl")
 export FinalJet, FinalJets
 
+# Substructure modules
+include("Substructure.jl")
```
graeme-a-stewart, line 93: Just for internal ordering, this should be after the `include("GenericAlgo.jl")` at L70, coming before the utility parts of the package.

src/Substructure.jl
```diff
@@ -0,0 +1,313 @@
...
+# Arguments
+- `jet`: The jet whose constituents are to be reclustered.
+- `clusterseq`: The cluster sequence from which the original jet is obtained.
+- `rad`: The new jet radius.
+- `mtd`: The new clustering method.
+
+# Returns
+- ClusterSequence: The new cluster sequence.
+"""
+function recluster(jet::PseudoJet, clusterseq::ClusterSequence, rad = 1.0, mtd = 0)
```
graeme-a-stewart, line 72: This function should have the same signature as `jet_reconstruct`. In particular the use of the power value as an alias for the algorithm I want to deprecate (even internally!); and the additional parameters should be named.

Also, calling the parameter `mtd` is inconsistent with the rest of the codebase, so rename to `p`. Ditto for `rad`, which should be `R`.

src/Substructure.jl
```diff
@@ -0,0 +1,313 @@
...
+    cons = constituents(jet, clusterseq)
+    new_clusterseq = jet_reconstruct(cons; p = mtd, R = rad, strategy = RecoStrategy.N2Plain)
+   
+    return new_clusterseq
+end
+
+function sort_jets!(event_jet_array)
+    jet_pt(jet) = pt2(jet)
+    sort!(event_jet_array, by = jet_pt, rev = true)
+    # println(event_jet_array)
```
graeme-a-stewart, line 82: Looks like it was for debugging, so remove

src/Substructure.jl
```diff
@@ -0,0 +1,313 @@
...
+end
+
+function join(jets::Vector{PseudoJet})
+    final = PseudoJet(0.0, 0.0, 0.0, 0.0)
+    
+    for jet in jets
```
graeme-a-stewart, line 88: You can replace with the more Julia idiomatic:

```Julia
foldl(+, jets)
```

Then you don't even need `final` I think.

In fact it probably doesn't even need the function wrapper now!

src/Substructure.jl
```diff
@@ -0,0 +1,313 @@
...
+Arguments:
+- jet: PseudoJet instance representing the jet to tag
+- clusterseq: ClusterSequence with jet clustering history
+- tag: MassDropTagger instance providing mass drop parameters
+
+Returns:
+- PseudoJet: The jet (or subjet) satisfying the mass drop conditions, if tagging is successful, otherwise a zero-momentum PseudoJet
+"""
+
+function massDrop(jet::PseudoJet, clusterseq::ClusterSequence, tag::MassDropTagger)
```
graeme-a-stewart, line 173: `massdrop`

src/Substructure.jl
```diff
@@ -0,0 +1,313 @@
...
+- jet: PseudoJet instance to groom
+- clusterseq: ClusterSequence containing jet history
+- rad: Radius parameter for reclustering
+- tag: SoftDropTagger instance with soft-drop parameters
+
+Returns:
+- PseudoJet: Groomed jet or zero-momentum PseudoJet if grooming fails
+"""
+
+function softDrop(jet::PseudoJet, clusterseq::ClusterSequence, rad::Float64, tag::SoftDropTagger)
```
graeme-a-stewart, line 219: `softdrop`

graeme-a-stewart: `rad` -> `R`

Should `R` also be part of the `SoftDropTagger` structure?

src/Substructure.jl
```diff
@@ -0,0 +1,313 @@
...
+
+Filters jets based on radius and number of hardest subjets, reducing contamination.
+
+Fields:
+- filterRadius: Radius parameter to recluster subjets
+- numHardestJets: Number of hardest jets to retain in the filtered result
+"""
+
+mutable struct Filter
+    filterRadius::Float64
```
graeme-a-stewart, line 137: Avoid camel case for variable names in this module - I think `R` is the only exception (mea culpa). So `filter_radius` and `num_hardest_jets`.

src/Substructure.jl
```diff
@@ -0,0 +1,313 @@
...
+
+Trims soft, large-angle components from jets based on fraction and radius.
+
+Fields:
+- trimRadius: Radius used for reclustering in trimming
+- trimFraction: Minimum momentum fraction for retained subjets
+- reclusterMethod: Method identifier for reclustering
+"""
+
+mutable struct Trim
```
graeme-a-stewart, line 152: Same comment for variable name - prefer lowercase and `_`s if needed.

src/Substructure.jl
```diff
@@ -0,0 +1,313 @@
...
+"""
+
+function softDrop(jet::PseudoJet, clusterseq::ClusterSequence, rad::Float64, tag::SoftDropTagger)
+    new_clusterseq = recluster(jet, clusterseq, rad, 0)
+    new_jet = inclusive_jets(new_clusterseq; T = PseudoJet)[1]
+
+    allJets = new_clusterseq.jets
+    hist = new_clusterseq.history
+    
+    while(true)
```
graeme-a-stewart, line 226: Don't use parentheses here: `while true` is sufficient

src/Substructure.jl
```diff
@@ -0,0 +1,313 @@
...
+
+Returns:
+- PseudoJet: The jet (or subjet) satisfying the mass drop conditions, if tagging is successful, otherwise a zero-momentum PseudoJet
+"""
+
+function massDrop(jet::PseudoJet, clusterseq::ClusterSequence, tag::MassDropTagger)
+    allJets = clusterseq.jets
+    hist = clusterseq.history
+
+    while(true)
```
graeme-a-stewart, line 177: Don't use unnecessary parentheses: `while true` is good enough.

https://docs.julialang.org/en/v1/manual/style-guide/#Don't-parenthesize-conditions

src/Substructure.jl
```diff
@@ -0,0 +1,313 @@
...
+        had_parents, p1, p2 = has_parents(jet, hist)
+
+        if (had_parents)
```
graeme-a-stewart, line 180: Ditto for unneeded parentheses

src/Substructure.jl
```diff
@@ -0,0 +1,313 @@
...
+            parent1 = allJets[hist[p1].jetp_index]
+            parent2 = allJets[hist[p2].jetp_index]
+
+            if (m2(parent1) < m2(parent2))
```
graeme-a-stewart, line 184: Ditto for unneeded parentheses...

(Largely I stop commenting on this now, but there are many cases)

src/Substructure.jl
```diff
@@ -0,0 +1,313 @@
...
+Arguments:
+- jet: PseudoJet instance representing the jet to filter
+- clusterseq: ClusterSequence containing jet history
+- filter: Filter instance specifying radius and number of subjets
+
+Returns:
+- PseudoJet: Filtered jet composed of the hardest subjets
+
+"""
+function jetFilter(jet::PseudoJet, clusterseq::ClusterSequence, filter::Filter)
```
graeme-a-stewart, line 269: `jetfilter` or `jet_filter`

src/Substructure.jl
```diff
@@ -0,0 +1,313 @@
...
+Arguments:
+- jet: PseudoJet instance representing the jet to trim
+- clusterseq: ClusterSequence containing jet history
+- trim: Trim instance specifying trimming parameters
+
+Returns:
+- PseudoJet: Trimmed jet composed of retained subjets
+"""
+
+function jetTrim(jet::PseudoJet, clusterseq::ClusterSequence, trim::Trim)
```
graeme-a-stewart, line 296: `jettrim`

src/Substructure.jl
```diff
@@ -0,0 +1,313 @@
...
+    hard = Vector{PseudoJet}(undef, 0)
+    for item in reclustered
+        if pt2(item) >= frac2 * pt2(jet)
+            push!(hard, item)
+        end
+    end
+    trimmed = join(hard)
+    
+    trimmed    
+end;
```
graeme-a-stewart, line 313: Semicolon not needed

graeme-a-stewart (changes requested):
Great - thank you @sattwamo. This looks pretty solid.

I have left a bunch of comments, which are mostly about coding style and being consistent with the conventions in the rest of the module.

graeme-a-stewart (changes requested):

src/Substructure.jl
graeme-a-stewart: Actually, maybe not, if this functionality is only for $pp$ events then it might be better to have the more restricted type...

graeme-a-stewart: And I would be in favour of removing the `_` from the function name and maybe moving it to a different source file

sattwamo: The function `_tj_dist(jet1, jet2)` seems to work only with `TiledJet` objects and is incompatible with `PseudoJet` type jets.

graeme-a-stewart:
Hello @sattwamo!

Sorry it took some time to get back to this as I was travelling. More detailed comments to come, but a few quick things:

## Tests

### Input files

Can you `gzip` the reference files for the tests? This keeps them small for the package. As they are JSON lists of numbers they compress really, really well:

```sh
 ~/code/juliadev/JetReconstruction/test/data/ [substructure] ls -l fastjet*
-rw-r--r--  1 graemes  staff  196740 Nov 20 15:24 fastjet-jet-filter.json
-rw-r--r--  1 graemes  staff  184888 Nov 20 15:24 fastjet-jet-trim.json
-rw-r--r--  1 graemes  staff  189687 Nov 20 15:24 fastjet-mass-drop.json
-rw-r--r--  1 graemes  staff  158266 Nov 20 15:24 fastjet-soft-drop.json
 ~/code/juliadev/JetReconstruction/test/data/ [substructure] gzip -9 fastjet*
 ~/code/juliadev/JetReconstruction/test/data/ [substructure*] ls -l fastjet*  
-rw-r--r--  1 graemes  staff  39751 Nov 20 15:24 fastjet-jet-filter.json.gz
-rw-r--r--  1 graemes  staff  30918 Nov 20 15:24 fastjet-jet-trim.json.gz
-rw-r--r--  1 graemes  staff  34650 Nov 20 15:24 fastjet-mass-drop.json.gz
-rw-r--r--  1 graemes  staff  17865 Nov 20 15:24 fastjet-soft-drop.json.gz
```

The `read_fastjet_outputs` will automatically decompress them for the test.

## Documentation

### Structures

You don't need, in fact you shouldn't, document the structures in `substructure.md`. The documentation you have written in `Substructure.jl` will be built and put into either the Public or Internal API pages automatically.

You reference it like this:

```
[`MassDropTagger`](@ref)
```

(there are a lot of examples in `index.md`).

What goes in `substructure.md` should be more descriptive of how to achieve an outcome, together with code snippets, and appropriate links to the API docs and examples (see, e.g., `EDM4hep.md`).

Overall things are looking in rather good shape now!

docs/src/substructure.md
```diff
@@ -0,0 +1,129 @@
+# Jet Substructure
+
+## Structures
```
graeme-a-stewart, line 3: Don't just repeat the docstrings! Write a description of how to use these functions and refer to the API documentation by reference (e.g. `[MassDropTagger](@ref)`).

You can refer to the nice examples you wrote as well.

docs/src/substructure.md
```diff
@@ -0,0 +1,129 @@
...
+
+**Fields**:
+
+- `trim_radius::Float64`: Radius used for reclustering in trimming.
+- `trim_fraction::Float64`: Minimum momentum fraction for retained subjets.
+- `recluster_method::JetAlgorithm.Algorithm`: Method identifier for reclustering.
+
+---
+
+## Functions
```
graeme-a-stewart, line 51: Same comment as above - don't repeat docstrings, show how to use these functions.

src/Substructure.jl
```diff
@@ -0,0 +1,304 @@
...
+
+    d_y = y1 - y2
+    d_phi = phi1 - phi2
+    d_phi = abs(d_phi) > π ? 2π - abs(d_phi) : d_phi
+
+    return sqrt(d_y^2 + d_phi^2)
+end
+
+# Function to calculate kt distance between two pseudojets
+function kt_distance(jet1::PseudoJet, jet2::PseudoJet)
```
graeme-a-stewart, line 47: Would be better with a proper docstring

src/Substructure.jl
```diff
@@ -0,0 +1,304 @@
...
+function recluster(jet::PseudoJet, clusterseq::ClusterSequence; R = 1.0,
+                   algorithm = JetAlgorithm.CA)
+    cons = constituents(jet, clusterseq)
+    new_clusterseq = jet_reconstruct(cons; p = nothing, R = R, algorithm = algorithm,
+                                     strategy = RecoStrategy.Best)
+
+    return new_clusterseq
+end
+
+function sort_jets!(event_jet_array)
```
graeme-a-stewart, line 78: I still find the need for this function questionable. You can just replace it with 

`sort!(event_jet_array, by = pt2, rev = true)`

so it's really not needed.

src/Substructure.jl
```diff
@@ -0,0 +1,304 @@
...
+    jet_pt(jet) = pt2(jet)
+    sort!(event_jet_array, by = jet_pt, rev = true)
+end
+
+# Defining suitable structures to be used for jet grooming and tagging
```
graeme-a-stewart, line 83: You can cut this line - say everything in the doctoring

src/Substructure.jl
```diff
@@ -0,0 +1,304 @@
...
+"""
+    struct MassDropTagger
+
+Used for tagging jets that undergo mass drop, a common technique in jet substructure.
+
+Fields:
+- mu: Maximum allowed mass ratio for a jet to pass tagging
+- y: Minimum kT distance threshold for parent separation
+"""
+
```
graeme-a-stewart, line 93: Remove blank line. (It may even cause trouble?)

src/Substructure.jl
```diff
@@ -0,0 +1,304 @@
...
+"""
+    struct JetFilter
+
+Filters jets based on radius and number of hardest subjets, reducing contamination.
+
+Fields:
+- filter_radius: Radius parameter to recluster subjets
+- num_hardest_jets: Number of hardest jets to retain in the filtered result
+"""
+
```
graeme-a-stewart, line 126: Remove blank line

src/Substructure.jl
```diff
@@ -0,0 +1,304 @@
...
+    struct JetTrim
+
+Trims soft, large-angle components from jets based on fraction and radius.
+
+Fields:
+- trim_radius: Radius used for reclustering in trimming
+- trim_fraction: Minimum momentum fraction for retained subjets
+- recluster_method: Method identifier for reclustering
+"""
+
```
graeme-a-stewart, line 142: Remove blank line

src/Substructure.jl
```diff
@@ -0,0 +1,304 @@
...
+Arguments:
+- jet: PseudoJet instance representing the jet to tag
+- clusterseq: ClusterSequence with jet clustering history
+- tag: MassDropTagger instance providing mass drop parameters
+
+Returns:
+- PseudoJet: The jet (or subjet) satisfying the mass drop conditions, if tagging is successful, otherwise a zero-momentum PseudoJet
+"""
+
```
graeme-a-stewart, line 163: Remove blank line

src/Substructure.jl
```diff
@@ -0,0 +1,304 @@
...
+Arguments:
+- jet: PseudoJet instance to groom
+- clusterseq: ClusterSequence containing jet history
+- tag: SoftDropTagger instance with soft-drop parameters
+
+Returns:
+- PseudoJet: Groomed jet or zero-momentum PseudoJet if grooming fails
+"""
+
```
graeme-a-stewart, line 207: Remove blank line

src/Substructure.jl
```diff
@@ -0,0 +1,304 @@
...
+        if had_parents
+            parent1 = all_jets[hist[p1].jetp_index]
+            parent2 = all_jets[hist[p2].jetp_index]
+
+            if m2(parent1) < m2(parent2)
+                p1, p2 = p2, p1
+                parent1, parent2 = parent2, parent1
+            end
+
+            pti = pt2(parent1)^0.5
```
graeme-a-stewart, line 229: Use `pt(parent1)` instead - it's defined for `PseudoJet`s

src/Substructure.jl
```diff
@@ -0,0 +1,304 @@
...
+            ptj = pt2(parent2)^0.5
```
graeme-a-stewart, line 230: Same

src/Substructure.jl
```diff
@@ -0,0 +1,304 @@
...
+
+            if min(pti, ptj) / (pti + ptj) >
+               tag.zcut * (deltaR(parent1, parent2) / rad)^tag.b
+                return new_jet
+            else
+                new_jet = parent1
+            end
+
+        else
+            return PseudoJet(0.0, 0.0, 0.0, 0.0)
```
graeme-a-stewart, line 240: I am not completely convinced by this return type - it might be better in this case to return `nothing`. But we need to check this is ok from a performance point of view. Could you check that?

src/Substructure.jl
```diff
@@ -0,0 +1,304 @@
...
+
+Arguments:
+- jet: PseudoJet instance representing the jet to trim
+- clusterseq: ClusterSequence containing jet history
+- trim: Trim instance specifying trimming parameters
+
+Returns:
+- PseudoJet: Trimmed jet composed of retained subjets
+"""
+
```
graeme-a-stewart, line 286: Remove blank

graeme-a-stewart (changes requested):
Only small changes needed now. Looking good!

src/Substructure.jl
```diff
@@ -0,0 +1,285 @@
+"""
+    has_parents(p, clusterseq) -> (boolean, Int64, Int64)
+
+Checks if the jet `p` is a child of two other jets, after clustering 
+
+# Arguments
+- `p`: The jet to check.
```
graeme-a-stewart, line 7: A small documentation point: the docstring should match the argument in the signature, including type. i.e., here it should be `p:: PseudoJet`.

src/Substructure.jl
```diff
@@ -0,0 +1,285 @@
...
+    return (result, p1, p2)
+end
+
+"""
+    deltaR(jet1, jet2) -> Float64
+
+Function to calculate the distance in the y-ϕ plane between two jets `jet1` and `jet2`
+
+# Arguments
+- `jet1`: The first jet.
```
graeme-a-stewart, line 29: `jet1::PseudoJet` etc.

src/Substructure.jl
```diff
@@ -0,0 +1,285 @@
...
+    return new_clusterseq
+end
+
+"""
+    struct MassDropTagger
+
+Used for tagging jets that undergo mass drop, a common technique in jet substructure.
+
+# Fields:
+- mu: Maximum allowed mass ratio for a jet to pass tagging
```
graeme-a-stewart, line 75: Needs the type and should be in "code" style, i.e.,

```md
- `mu::Float64`: ...
```

graeme-a-stewart (changes requested):

examples/Substructure/JetGrooming.jl
```diff
@@ -0,0 +1,39 @@
+#! /usr/bin/env julia
```
graeme-a-stewart, line 1: One small additional point - favour lowercase for directory names here, so make the directory `substructure`.


Yes, `EDM4hep` breaks this rule, but it's because EDM = Event Data Model, so it's a acronym; plus it's the actual name of the package.

graeme-a-stewart (changes requested):

graeme-a-stewart:
> Great @sattwamo - all looks good to me bar a small improvement in the docstrings. Can you make sure that in the docstring for functions and structs the arguments or fields are:
> 
> * Formatted as `code` with ``
> * Have the types present, if these are specified, i.e., not just `mu` but `mu::Float64`
> 
> I pointed out a few of these cases, but not all, in code comments.

To finesse this, if the function signature is short and the variable names are clear then it's not necessary to list them separately in the docstring (thanks to @grasph for pointing out this out: https://docs.julialang.org/en/v1/manual/documentation/#Writing-Documentation).

Do make sure that type specifiers are present!

graeme-a-stewart:
Almost perfect - we just need to rename the examples directory as `examples/substructure`, *lower case*.

This can be a bit tricky if you are using git on a case insensitive filesystem (which is the default on OS X), but it can be managed - it's better still to fix this on this branch, because it can cause some issues for OS X if case changes between different commits in the same repository.

Let me know if you want some help to do this bit.

graeme-a-stewart (approved):

graeme-a-stewart:
Now merged - thank you very much @sattwamo!

PR #84, opened 2024-10-27
use SIMD.jl directly instead of LV.jl for `fast_findmin()`

Moelf:
fix #83

PR #88, opened 2024-11-11
Add submodule for C bindings, static compilation

m-fila:
Adding C bindings and ahead of time compilation to shared library with PackageCompiler.jl or juliac

C components have `jetreconstruction_`(or `JETRECONSTRUCTION_`) prefix

For now the C API includes only:
- `jet_reconstruct`
- `exclusive_jets`, `inclusive_jets`
- enums and structs used in their signatures. When needed new structs with C-like layout are introduced (for instance replacing `Vector{T}` field with `Ptr{T}` and `Csize_t` fields for C-facing structs)

The C header file has to be written manually and keep in sync with Julia code. Technically the whole `compile` could be a separate julia package possibly even living in a separate repo, but I think it'd be harder to keep in sync this way
 

---
To do:
- [x]  write to an output pointer from `inclusive_jets` and `exclusive_jets`
- [x] add error handling
- [x] add missing docstrings
- [x] add precompilation statements
- [x] cache all combinations of strategy and algorithm in precompilation
- [x] have actual test
- [x] install `JetReconstructionConfig.cmake`
- [x] compilation with `juliac`
- [x] add GenKt
- [x] add recombination schemas
- [ ] support Lorentz vectors
- [ ] support EEJet


m-fila:
The CI for Julia 1.9 fails as documenting`@ccallable` functions was fixed only in 1.11 :cry: 

I couldn't think of a good way of error handling.
What I propose here is to:
- use status codes from enum with:
   - value for success,
   - values for each standard Julia exception,
   - single value for the remaining cases.
- print the exception and backtrace to stderr so it's not lost

I also couldn't find a reasonable way to translate Julia exceptions to C++ exceptions without going through C-bottleneck

graeme-a-stewart:
Yeah, I would agree we can drop support for 1.9 now, but the LTS release (1.10) we should try to support

m-fila:
Thanks for checking. I think that was because the `results_ptr`pointee was managed by Julia and got garbage collected. Should be fixed now

graeme-a-stewart:
Yes, now it works - thanks for the quick fix @m-fila!

m-fila:
Added compiling with `juliac` from Julia nightlies. Unfortunately `LoopVectorization` isn't compatible with `juliac` and would have to be removed altogether from the project.
Also any kind of trimming produces a lot of errors and aborts the compilation.

With the nightlies I get a lot of warnings like this. They might be false positives though:
```
WARNING: Detected access to binding `JetReconstruction.##meta#64` in a world prior to its definition world.
  Julia 1.12 has introduced more strict world age semantics for global bindings.
  !!! This code may malfunction under Revise.
  !!! This code will error in future versions of Julia.
Hint: Add an appropriate `invokelatest` around the access to this binding.
```

Taking all into account some results

|  | PackageCompiler | juliac |
|-|-|-|
| Compilation time (s)| ~730 | ~90 |
| Lib size (M) | 283 | 247 |
| Deps size (M) | 279 | 0 |

with `juliac` the explicit initialization and shutdown of julia runtime is gone. The pre-compilation is also gone (I didn't manage to reduce completely the overheads of first execution with PackageCompiler either with precompilation-statements, precompilations-execution or precompilation-execution on the test suite). Also now the compiled object doesn't bring any exclusive dependencies that should be kept around it.

Moelf:
sounds good! 

>Added compiling with juliac from Julia nightlies. Unfortunately LoopVectorization isn't compatible with juliac and would have to be removed altogether from the project.

for this, I guess time to retry https://github.com/JuliaHEP/JetReconstruction.jl/pull/84

m-fila:
> for this, I guess time to retry #84

Thankfully SIMD and its deps seem to be working fine with juliac

m-fila:
`PackageCompiler` with `LoopVectorization` compiles and runs but somehow is much slower ( more than x5) even than than a simple loop with no extra optimization

Moelf:
https://discourse.julialang.org/t/guide-using-sum-types-to-define-dynamic-c-apis-for-julia-1-12-trim/132111

seems applicable

m-fila:
I think we don't have anything like this here. From what I understood their function was type-unstable from the begging as returned type was decided by enum which isn't very nice even without juliac involved.

We use enums to select which function (strategy, algorithm) but it's stable as in the end we get the same type, if a function can select return type (jet selection) we do so by passing `Type` explicitly no enums involved.
In the C-interface here we only have single jet type `PseudoJet`.

m-fila:
Superseded by #242

PR #98, opened 2024-11-27
Improved constituents

### LINKED ISSUE #68: Support retrieval of jet constituents

graeme-a-stewart:
Formally added functions now to retrieve jet constituents - either as a list of indexes to the original particle collection, or as references to the reconstructed jet objects.

Documentation and examples are added.

Closes #68

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -2,7 +2,7 @@
 
 ## Adding Documentation
 
-`Documenter.jl` is used to generate package documentation. 
+`Documenter.jl` is used to generate package documentation.
 
 The general guidelines are to have documentation split into meaningful sections,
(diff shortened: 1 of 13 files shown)


mgratrix:
JetReconstruction.constituents works incredibly well at extracting constituents of each jet. I have yet to face any complications!

PR #114, opened 2025-01-28
Add Aqua.jl tests

graeme-a-stewart:
Enable testing with Aqua.jl (https://juliatesting.github.io/Aqua.jl/stable/).

To enable this, move to a separate Project.toml for the tests, to cleanly separate the environments. Add a README to explain basic layout and execution of tests.

Remove the unneeded JSON dependency from the main project.

Add some compatibilities for some standard library packages.

Fix a type piracy on `tryparse` and optimise parsing of enum values from strings.

PR #115, opened 2025-01-29
Activate the test env with TestEnv.jl

m-fila:
This is a follow-up from #114

- Adding Agua badge in readme and putting aqua tests in a testset
- Merging back `test/Project.toml` with `Project.toml`, and using TestEnv.jl to activate the test environment

TestEnv can be used to programmatically activate a test environment so there is no need to create a separate project for tests.
Running separate files can be done from shell with the main project which I think is quite nice and simple:

```
julia --project test/test-pp-reconstruction.jl
```

The cost of this is that the `TestEnv` should be installed in the default environment, which should be fine since it's small and self-contained, and [other packages](https://github.com/LuxDL/Lux.jl?tab=readme-ov-file#-testing) also suggest using it.

Anyway if TestEnv is too much then I can revert to the current solution with `test/Project.toml` and propose some changes for it.


graeme-a-stewart:
@all-contributors please add @m-fila for infrastructure, tests

PR #118, opened 2025-01-30
allow exclusive and inclusive jets to return jet type used in clusterseq

m-fila:
Fixing `exclusive_jets` and `inclusive_jets` so they can return either a `LorentzVectorCyl` or the type used for jets in input ClusterSequence.

Previously it wasn't possible to obtain `EEjet` from `inclusive_jets` .
For both inclusive and exclusive it was possible to try converting `ClusterSequence{EEjet}` to `Pseudojet` - now requesting wrong type will produce `LorentzVectorCyl`

### DIFF
--- a/src/ClusterSequence.jl
+++ b/src/ClusterSequence.jl
@@ -214,7 +214,7 @@ add_step_to_history!(clusterseq::ClusterSequence, parent1, parent2, jetp_index, 
 end
 
 """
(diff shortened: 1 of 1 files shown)


graeme-a-stewart (approved):
Thank you @m-fila - this was overlooked when the parametrised version of `ClusterSequence` was introduced.

PR #119, opened 2025-01-31
fix GenKt not recognized as pp algorithm

m-fila:
GenKt wasn't recognized as a pp algorithm and `jet_reconstruction` was failing to resolve which algorithm function to use in that case.
The tests for GenKt seemed to be missing

Reference file `jet-collections-fastjet-inclusive-genkt-p1.5.json.gz` was using different capitalization than files for the algorithms. I couldn't find anything relaying on the filename being spelled this way so renamed for consistency

There is also `do_test_compare_to_fastjet` function that seems to be unused in the current tests and is probably a relic from #73 
https://github.com/JuliaHEP/JetReconstruction.jl/blob/96bde1b8e093afddddd40c52396f516064434efe/test/runtests.jl#L66 Let me know if we could remove it


m-fila:
Currently `GenKt` can't be also used with `exclusive_jets` and `n_exclusive_jets`. I think there should be a check for power > 0 if GenKt is used, right?

https://github.com/JuliaHEP/JetReconstruction.jl/blob/96bde1b8e093afddddd40c52396f516064434efe/src/ClusterSequence.jl#L323-L327

https://github.com/JuliaHEP/JetReconstruction.jl/blob/96bde1b8e093afddddd40c52396f516064434efe/src/ClusterSequence.jl#L380-L384

graeme-a-stewart:
Hello @m-fila 

Thanks for this - when I generalised the test setup for e+e- in #73 the GenKt test got lost!

Also, the `do_test_compare_to_fastjet` is indeed and anachronism now, so it can be deleted.

test/test-pp-reconstruction.jl
```diff
@@ -4,14 +4,22 @@
...
 
 # Test inclusive jets for each algorithm and strategy
-for alg in [JetAlgorithm.AntiKt, JetAlgorithm.CA, JetAlgorithm.Kt],
+for alg in [JetAlgorithm.AntiKt, JetAlgorithm.CA, JetAlgorithm.Kt, JetAlgorithm.GenKt],
     stg in [RecoStrategy.N2Plain, RecoStrategy.N2Tiled]
 
-    fastjet_file = joinpath(@__DIR__, "data",
-                            "jet-collections-fastjet-inclusive-$(alg).json.gz")
+    if alg == JetAlgorithm.GenKt
+        power = 1.5
```
graeme-a-stewart, line 11: I am just thinking it would be better to define something like:

```julia
test_genkt_power = 1.4
test_cone_size=0.4
```

Then here we'd have `power = test_genkt_power`...

test/test-pp-reconstruction.jl
```diff
@@ -4,14 +4,22 @@
...
+    else
+        power = JetReconstruction.algorithm2power[alg]
+        fastjet_file = joinpath(@__DIR__, "data",
+                                "jet-collections-fastjet-inclusive-$(alg).json.gz")
+    end
+
     test = ComparisonTest(events_file_pp, fastjet_file,
                           alg, stg,
-                          JetReconstruction.algorithm2power[alg], 0.4,
+                          power, 0.4,
```
graeme-a-stewart, line 22: ...and here `test_cone_size`

graeme-a-stewart (changes requested):
Just a small additional improvement suggested

test/test-pp-reconstruction.jl
m-fila: Thanks, done

src/ClusterSequence.jl
```diff
@@ -378,8 +380,10 @@ n_exclusive_jets(clusterseq, dcut = 20.0)
 """
 function n_exclusive_jets(clusterseq::ClusterSequence; dcut::AbstractFloat)
     # Check that an algorithm was used that makes sense for exclusive jets
-    if !(clusterseq.algorithm ∈
-         (JetAlgorithm.CA, JetAlgorithm.Kt, JetAlgorithm.EEKt, JetAlgorithm.Durham))
+    if clusterseq.algorithm == JetAlgorithm.GenKt && clusterseq.power < 0
+        throw(ArgumentError("Algorithm GenKt requires power >= 0 (power=$(clusterseq.power))"))
+    elseif clusterseq.algorithm ∉
+           (JetAlgorithm.CA, JetAlgorithm.Kt, JetAlgorithm.EEKt, JetAlgorithm.Durham)
```
graeme-a-stewart, line 386: There is a logic flaw here! If you have GenKt and p >= 0, this test will *fail* and it should pass.

```sh
 ~/.julia/dev/JetReconstruction/examples/ [genkt_pp] julia --project jetreco.jl -p 0.5 -A GenKt -R 1.0 --exclusive-njets 2 ../test/data/events.pp13TeV.hepmc3.gz
[ Info: Total Events: 100
[ Info: Jet reconstruction will use GenKt with power 0.5
[ Info: Will produce exclusive jets with n_jets = 2
ERROR: LoadError: ArgumentError: Algorithm used is not suitable for exclusive jets (GenKt)
...
```

src/ClusterSequence.jl
```diff
@@ -321,8 +321,10 @@ function exclusive_jets(clusterseq::ClusterSequence{U}; dcut = nothing, njets =
     end
 
     # Check that an algorithm was used that makes sense for exclusive jets
-    if !(clusterseq.algorithm ∈
-         (JetAlgorithm.CA, JetAlgorithm.Kt, JetAlgorithm.EEKt, JetAlgorithm.Durham))
+    if clusterseq.algorithm == JetAlgorithm.GenKt && clusterseq.power < 0
+        throw(ArgumentError("Algorithm GenKt requires power >= 0 (power=$(clusterseq.power))"))
+    elseif clusterseq.algorithm ∉
+           (JetAlgorithm.CA, JetAlgorithm.Kt, JetAlgorithm.EEKt, JetAlgorithm.Durham)
```
graeme-a-stewart, line 327: Same logic flaw as below - GenKt with p >= 0 will fail this test and throw an error

graeme-a-stewart (changes requested):

graeme-a-stewart:
Hi @m-fila - there's a logic flaw now that prevents GenKt with p >= 0 from running exclusive selections.

I also realised that the same argument applied to EEKt - it should be tested for p >= 0.

So the logic should be more like:

```Julia
    # Check that an algorithm was used that makes sense for exclusive jets
    if (clusterseq.algorithm ∈ (JetAlgorithm.GenKt, JetAlgorithm.EEKt)) && clusterseq.power < 0
        throw(ArgumentError("Algorithm $(clusterseq.algorithm) requires power >= 0 (power=$(clusterseq.power))"))
    elseif clusterseq.algorithm ∉
           (JetAlgorithm.CA, JetAlgorithm.Kt, JetAlgorithm.Durham, JetAlgorithm.GenKt, JetAlgorithm.EEKt)
        throw(ArgumentError("Algorithm used is not suitable for exclusive jets ($(clusterseq.algorithm))"))
    end
```

graeme-a-stewart:
It might be useful to write a few tests against the wrong setup, checking exceptions are thrown correctly (would help codecov!).

src/ClusterSequence.jl
```diff
@@ -321,8 +321,10 @@ function exclusive_jets(clusterseq::ClusterSequence{U}; dcut = nothing, njets =
     end
 
     # Check that an algorithm was used that makes sense for exclusive jets
-    if !(clusterseq.algorithm ∈
-         (JetAlgorithm.CA, JetAlgorithm.Kt, JetAlgorithm.EEKt, JetAlgorithm.Durham))
+    if clusterseq.algorithm == JetAlgorithm.GenKt && clusterseq.power < 0
+        throw(ArgumentError("Algorithm GenKt requires power >= 0 (power=$(clusterseq.power))"))
```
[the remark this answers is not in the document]
m-fila: Right, changed to `"Algorithm $(clusterseq.algorithm) requires power >= 0 for exclusive jets (power=$(clusterseq.power))"`

m-fila:
Thanks, fixed message and logic. Added standalone file with tests - I didn't like splitting them and putting in existing pp and ee tests

test/test-selection.jl
```diff
@@ -0,0 +1,46 @@
+# File: test/test_test-selection.jl
```
graeme-a-stewart, line 1: Filename is wrong.

The comment should describe better the aim of these tests of exclusive jet selection

graeme-a-stewart (changes requested):
Just one tiny little snag - the comment at the top of the test-selection file. Otherwise LGTM!

test/test-selection.jl
m-fila: Sorry, must be some stupid IDE extension :sweat_smile:

graeme-a-stewart (approved):

graeme-a-stewart:
Looks good to me - thanks! 🙏

PR #128, opened 2025-02-28
Fix default p value and refactor tests

### LINKED ISSUE #126: Unspecified power default value is wrong in `jet_reconstruct` - breaks CA and Kt reco
The current signature for `jet_reconstruct` (the generic entry point) is:

```Julia
function jet_reconstruct(particles; p::Union{Real, Nothing} = -1, R = 1.0,
                         algorithm::Union{JetAlgorithm.Algorithm, Nothing} = nothing,
                         recombine = +,
                         strategy = RecoStrategy.Best)
```

However this doesn't work as intended. If the user passes an `algorithm` value, but not a `p` value (which is currently recommended) then the `p` is set to -1.

If the algorithm is then not consistent with $p=-1$, i.e., Cambridge/Aachen or Inclusive Kt then the power and algorithm are now inconsistent and an exception is thrown.

The default value for `p` should be `nothing`.

### LINKED ISSUE #127: Complete refactoring of tests into subtests
There are a few tests still run directly in `runtests.jl`:

- algorithm/power consistency checks
- input data type consistency checks

These should be factored out into stand alone subtests so that they can be run independently and `runtests.jl` is just the wrapper that executes all subtests.

graeme-a-stewart:
Fix the default value for `p` in `jet_reconstruct()` to be `nothing` (see #126).

Refactor the last tests run inside `runtests.jl` to be independent subtests.

Add a couple more example test cases 

Fixes #126 
Fixes #127

### DIFF
--- a/test/test-jet_reconstruct-if.jl
+++ b/test/test-jet_reconstruct-if.jl
@@ -5,25 +5,43 @@ include("common.jl")
 let inputs = JetReconstruction.read_final_state_particles(events_file_ee)
     @testset "jet_reconstruct() interface" begin
         # EE Algorithms
-        @test typeof(jet_reconstruct(inputs[1]; algorithm = JetAlgorithm.Durham)) == ClusterSequence{EEjet}
-        @test typeof(jet_reconstruct(inputs[1]; algorithm = JetAlgorithm.EEKt, p=-1, R=1.0)) == ClusterSequence{EEjet}
-        @test_throws ArgumentError typeof(jet_reconstruct(inputs[1]; algorithm = JetAlgorithm.EEKt)) == ClusterSequence{EEjet}
+        @test typeof(jet_reconstruct(inputs[1]; algorithm = JetAlgorithm.Durham)) ==
+              ClusterSequence{EEjet}
+        @test typeof(jet_reconstruct(inputs[1]; algorithm = JetAlgorithm.EEKt, p = -1,
+                                     R = 1.0)) == ClusterSequence{EEjet}
+        @test_throws ArgumentError typeof(jet_reconstruct(inputs[1];
+                                                          algorithm = JetAlgorithm.EEKt))==ClusterSequence{EEjet}
 
         # PP Algorithms
-        @test typeof(jet_reconstruct(inputs[1]; algorithm = JetAlgorithm.AntiKt)) == ClusterSequence{PseudoJet}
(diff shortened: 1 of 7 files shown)


test/test-jet_reconstruct-if.jl
```diff
@@ -0,0 +1,47 @@
+# Test the different interface modes of the main jet_reconstruct() function
+
```
m-fila, line 2: Should this file also include tests for calling `jet_reconstuction` without algorithm argument since it's allowed? Something like?
```julia
@test typeof(jet_reconstruct(inputs[1]; p = 1.0,R = 0.4)) == ClusterSequence{PseudoJet}
```

m-fila (commented):
Maybe the file `test-jet_reconstruct-if.jl` could be renamed to use dash instead of underscore for consistency with other tests? Although I see the intent was to use `jet_reconstruct` there as a name of function.
Or probably renamed to `test-jet_reconstruct-interface.jl`?

I'm also wondering if at this point wouldn't it make sense to make `algorithm` argument mandatory and then handle `p` or go with https://github.com/JuliaHEP/JetReconstruction.jl/issues/126#issuecomment-2689265405. The default function could for instance throw or give `nothing` for `GenKt` or `EEKt`?

graeme-a-stewart:
> Maybe the file `test-jet_reconstruct-if.jl` could be renamed to use dash instead of underscore for consistency with other tests? Although I see the intent was to use `jet_reconstruct` there as a name of function. Or probably renamed to `test-jet_reconstruct-interface.jl`?

The `_` was exactly deliberate as it's the name of the method. Yes, `interface` would be clearer than `uf` 👍. 

> 
> I'm also wondering if at this point wouldn't it make sense to make `algorithm` argument mandatory and then handle `p` or go with [#126 (comment)](https://github.com/JuliaHEP/JetReconstruction.jl/issues/126#issuecomment-2689265405). The default function could for instance throw or give `nothing` for `GenKt` or `EEKt`?

I was considering this (see https://github.com/JuliaHEP/JetReconstruction.jl/blob/59f3f7032497c9ce099eb2d5149cef24b85a7a59/docs/src/index.md?plain=1#L62), but changing from an optional to a mandatory argument is a breaking change, so I would reserve that for 0.5.0.

m-fila (approved):
LGTM

graeme-a-stewart:
Useful comments as ever - thanks @m-fila! 🙏

PR #132, opened 2025-03-12
Support zstd compressed files

### LINKED ISSUE #131: Support zstd compression
ZStd gets much better compression than Gzip, but is also very fast to decompress (unlike LZMA), so it would be good to support it for decompressing

- ASCII HepMC files
- JSON comparison results from Fastjet

using https://github.com/JuliaIO/CodecZstd.jl

Migrating the actual data files we use is a different issue as this might break someone's workflow (but we never guaranteed these files' locations for all time).

graeme-a-stewart:
This PR makes a small enhancement to support reading Zstd compressed files (`*.zst`). Zstd can get 20% better compression than Gzip and reading is as fast (unlike LZMA).

File opening is abstracted using a small utility wrapper, so reduces code duplication.

Closes #131


src/Utils.jl
```diff
@@ -1,6 +1,20 @@
...
+using Match
+
+"""
+    open_with_stream(fname::AbstractString)
+
+Open a file with a stream decompressor if it is compressed with gzip or zstd,
+otherwise as a normal file.
+"""
+open_with_stream(fname::AbstractString) = @match fname begin
+    r"\.gz$" => GzipDecompressorStream(open(fname))
```
Moelf, line 14: do we need `@match`, or do we just want to check
```julia
endswith(fname, ".gz")
```

graeme-a-stewart: That is the way that the original code was, but I have an aversion to nested `if` `elseif` blocks when `@match` is more compact (at heart, I am hankering for `switch:`). This should be a very optimised regex when compiled, so I'm not worried about performance.

graeme-a-stewart: One may even consider adding more stream readers, thought it's probably fine with just these two

graeme-a-stewart: `Match` is a bit OTT for this, so I reimplemented with a few nested `if`s

PR #134, opened 2025-03-14
add mutable keyword in mutable struct docstrings

m-fila:
This is very small. The docstring for `PseudoJet` includes `mutable` but docstings for other `mutable struct`s don't. I'm not sure whether including `mutable` is a recommended practice - I think I've seen packages both putting and avoiding it in the docs

### DIFF
--- a/src/EEjet.jl
+++ b/src/EEjet.jl
@@ -1,5 +1,5 @@
 """
-    struct EEjet
+    mutable struct EEjet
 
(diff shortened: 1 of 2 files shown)


graeme-a-stewart (approved):
I think it's good to be accurate with this and say what the structure is. No one should ever use `TiledJet`, but `EEjet`s could be valid to construct in user code.

PR #138, opened 2025-03-24
Removed Jet Substructure Structs

sattwamo:
Using separate structs for calling substructure functions seemed redundant. Removed the structs and incorporated their functionalities into the functions itself.

### DIFF
--- a/docs/src/substructure.md
+++ b/docs/src/substructure.md
@@ -5,21 +5,20 @@ Jet substructure techniques provide powerful tools for analyzing and refining th
 ### Mass Drop Tagging
 
 ```julia
-mass_drop(jet::PseudoJet, clusterseq::ClusterSequence, tag::MassDropTagger) -> PseudoJet
+mass_drop(jet::PseudoJet, clusterseq::ClusterSequence; mu::Float64, y::Float64) -> PseudoJet
 ```
 
 The [mass_drop](@ref) function identifies subjets in a jet that pass the mass drop tagging condition. To use the `mass_drop` function:
 
-- Configure the [MassDropTagger](@ref) with parameters tailored to the expected mass drop and distance thresholds for analysis.
-
-```julia
-tagger = MassDropTagger(mu=0.67, y=0.09)
-```
-
 - Apply the `mass_drop` function to the jet and its clustering sequence.
 
 ```julia
-tagged_jet = mass_drop(jet, clusterseq, tagger)
+tagger = (mu=0.67, y=0.09)
+tagged_jet = mass_drop(jet, clusterseq; tagger...)
+```
+or
+```julia
+tagged_jet = mass_drop(jet, clusterseq; mu = 0.67, y = 0.09)
 ```
 
 - If the jet is tagged successfully, the function returns the identified subjet. Else it returns `PseudoJet(0.0, 0.0, 0.0, 0.0)`.
@@ -34,22 +33,25 @@ soft_drop(jet::PseudoJet, clusterseq::ClusterSequence, tag::SoftDropTagger) -> P
 
 The [soft_drop](@ref) function applies soft-drop grooming to remove soft, wide-angle radiation from jets. It reclusters the jet with a specified radius and clustering method, iteratively checking the soft-drop condition on subjets. To use the `soft_drop` function:
 
-- Construct the [SoftDropTagger](@ref) by defining the parameters for the grooming process, such as the energy fraction (`zcut`) and angular exponent (`b`)
-
-```julia
-tagger = SoftDropTagger(zcut=0.1, b=2.0)
-```
-
-By default, the reclustering radius is set to `1.0` which can be modified by defing the `tagger` object as
-
-```julia
-tagger = SoftDropTagger(zcut=0.1, b=2.0, cluster_rad=0.4)
-```
-
 - Apply the `soft_drop` function to the jet and its clustering sequence.
 
 ```julia
-tagged_jet = soft_drop(jet, clusterseq, tagger)
+tagger = (zcut = 0.1, beta = 2.0)
+tagged_jet = soft_drop(jet, clusterseq; tagger...)
+```
+or
+```julia
+tagged_jet = soft_drop(jet, clusterseq; zcut = 0.1, beta = 2.0)
+```
+
+By default, the reclustering radius is set to `1.0` which can be modified by the user as:
+```julia
+tagger = (zcut = 0.1, beta = 2.0, radius = 0.4)
+tagged_jet = soft_drop(jet, clusterseq; tagger...)
+```
+or
+```julia
+tagged_jet = soft_drop(jet, clusterseq; zcut = 0.1, beta = 2.0, radius = 0.4)
 ```
 
 - If the jet is tagged successfully, the function returns the identified subjet. Else it returns `nothing`.
@@ -64,16 +66,15 @@ jet_filtering(jet::PseudoJet, clusterseq::ClusterSequence, filter::JetFilter) ->
 
 The [jet_filtering](@ref) function filters a jet to retain only the hardest subjets based on a specified radius and number. To use the function:
 
-- Set the radius for subjet clustering and the number of hardest subjets to retain in a [JetFilter](@ref) method.
-
-```julia
-filter = JetFilter(filter_radius=0.3, num_hardest_jets=3)
-```
-
 - Apply the `jet_filtering` function to refine the jet.
 
 ```julia
-filtered_jet = jet_filtering(jet, clusterseq, filter)
+filter = (radius=0.3, hardest_jets=3)
+filtered_jet = jet_filtering(jet, clusterseq; filter...)
+```
+or
+```julia
+filtered_jet = jet_filtering(jet, clusterseq; radius = 0.3, hardest_jets = 3)
 ```
 
 - The function returns the filtered jet that retains only the most significant subjets, reducing noise.
@@ -88,21 +89,20 @@ jet_trimming(jet::PseudoJet, clusterseq::ClusterSequence, trim::JetTrim) -> Pseu
 
 The [jet_trimming](@ref) function trims a jet by removing subjets with transverse momentum below a specified fraction of the main jet's momentum. This method cleans up jets by removing soft particles. To use this function:
 
-- Configure the trimming radius and momentum fraction threshold in [JetTrim](@ref)
-
-```julia
-trim = JetTrim(trim_radius=0.3, trim_fraction=0.3, recluster_method=JetAlgorithm.CA)
-```
-
-It is to be noted that the `jet_trimming` function reclusters the constituents of the jet using either `C/A` or `kT ` algorithm, which needs to be specified.
-
 - Apply the `jet_trimming` function to clean the jet.
 
 ```julia
-trimmed_jet = jet_trimming(jet, clusterseq, trim)
+trim = (radius=0.3, fraction=0.3, recluster_method=JetAlgorithm.CA)
+trimmed_jet = jet_trimming(jet, clusterseq; trim...)
+```
+or
+```julia
+trimmed_jet = jet_trimming(jet, clusterseq; radius=0.3, fraction=0.3, recluster_method=JetAlgorithm.CA)
 ```
 
 - The function returns the trimmed jet.
+
+It is to be noted that the `jet_trimming` function reclusters the constituents of the jet using either `C/A` or `kT ` algorithm, which needs to be specified.
 
 ---
 
--- a/src/Substructure.jl
+++ b/src/Substructure.jl
@@ -45,69 +45,6 @@ function recluster(jet::PseudoJet, clusterseq::ClusterSequence; R = 1.0,
 end
 
 """
-    struct MassDropTagger
-
-Used for tagging jets that undergo mass drop, a common technique in jet substructure.
-
-# Fields:
-- `mu::Float64`: Maximum allowed mass ratio for a jet to pass tagging.
-- `y::Float64`: Minimum kT distance threshold for parent separation.
-"""
-struct MassDropTagger
-    mu::Float64
-    y::Float64
-end
-
-"""
-    struct SoftDropTagger
-
-Applies a soft-drop condition on jets, trimming away soft, wide-angle radiation.
-
-# Fields:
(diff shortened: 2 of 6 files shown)


m-fila:
Do we want to remove struct based function straight-away or deprecate and keep a little longer in parallel with the new functions?

graeme-a-stewart:
> @graeme-a-stewart Do we want to remove struct based function straight-away or deprecate and keep a little longer in parallel with the new functions?

My feeling is that we bite the bullet and remove them. We will break a few things in 0.5.0... People can keep using the 0.4.X series until they are ready to change - I am prepared to do further 0.4.X releases for fixing bugs, at least for a while.

They are a new feature, so I think it's better to clean up the signatures unambiguously ASAP.

src/Substructure.jl
```diff
@@ -116,12 +53,13 @@ The method stops at the first jet satisfying the mass and distance thresholds.
...
 - `clusterseq::ClusterSequence`: ClusterSequence with jet clustering history.
-- `tag::MassDropTagger`: MassDropTagger instance providing mass drop parameters.
+- `mu::Float64`: Maximum allowed mass ratio for a jet to pass tagging.
+- `y::Float64`: Minimum kT distance threshold for parent separation.
 
 # Returns:
 - `PseudoJet`: The jet (or subjet) satisfying the mass drop conditions, if tagging is successful, otherwise a zero-momentum PseudoJet
 """
-function mass_drop(jet::PseudoJet, clusterseq::ClusterSequence, tag::MassDropTagger)
+function mass_drop(jet::PseudoJet, clusterseq::ClusterSequence; mu::Float64, y::Float64)
```
m-fila, line 62:
```suggestion
function mass_drop(jet::PseudoJet, clusterseq::ClusterSequence{PseudoJet}; mu::Float64, y::Float64)
```

could we use this opportunity to also constrain `ClusterSequence` parameter here and and other substructure functions?

graeme-a-stewart:
@sattwamo - your PR is based on quite an old commit, from January. Could you *rebase* to the head of `main` please? It will make the change more consistent with the rest of the codebase. Thanks.

docs/src/substructure.md
```diff
@@ -5,21 +5,20 @@ Jet substructure techniques provide powerful tools for analyzing and refining th
...
+tagged_jet = mass_drop(jet, clusterseq; tagger...)
 ```
-
-- Apply the `mass_drop` function to the jet and its clustering sequence.
-
+or
 ```julia
-tagged_jet = mass_drop(jet, clusterseq, tagger)
+tagged_jet = mass_drop(jet, clusterseq; mu = 0.67, y = 0.09)
 ```
```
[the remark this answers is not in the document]
graeme-a-stewart: Yes, I agree - the function signature makes it clear how this works in case a user wants to do something non-standard here.

src/Substructure.jl
```diff
@@ -158,15 +96,17 @@ This function reclusters the jet and iteratively checks the soft-drop condition
 # Arguments:
 - `jet::PseudoJet`: PseudoJet instance to groom.
 - `clusterseq::ClusterSequence`: ClusterSequence containing jet history.
-- `tag::SoftDropTagger`: SoftDropTagger instance with soft-drop parameters.
+- `zcut::Float64`: Minimum allowed energy fraction for subjets.
```
graeme-a-stewart, line 99: Not here, but at L91, the docstring signature is wrong.

src/Substructure.jl
```diff
@@ -204,18 +144,18 @@ Filters a jet to retain only the hardest subjets based on a specified radius and
 # Arguments:
 - `jet::PseudoJet`: PseudoJet instance representing the jet to filter.
 - `clusterseq::ClusterSequence`: ClusterSequence containing jet history.
-- `filter::JetFilter`: Filter instance specifying radius and number of subjets.
+- `radius::Float64`: Radius parameter to recluster subjets.
```
graeme-a-stewart, line 147: At L140 the function signature is wrong.

src/Substructure.jl
```diff
@@ -231,17 +171,18 @@ Trims a jet by removing subjets with transverse momentum below a specified fract
 # Arguments:
 - `jet::PseudoJet`: PseudoJet instance representing the jet to trim.
 - `clusterseq::ClusterSequence`: ClusterSequence containing jet history.
-- `trim::JetTrim`: Trim instance specifying trimming parameters.
+- `radius::Float64`: Radius used for reclustering in trimming.
```
graeme-a-stewart, line 174: At L167 the function signature is wrong

src/Substructure.jl
graeme-a-stewart, line 62: There is no need to force `mu` and `y` to be `Float64` here - put `Real`.

src/Substructure.jl
```diff
@@ -158,15 +96,17 @@ This function reclusters the jet and iteratively checks the soft-drop condition
...
 # Returns:
-- `PseudoJet`: Groomed jet or zero-momentum PseudoJet if grooming fails.
+- `PseudoJet`: Groomed jet or `nothing` if grooming fails.
 """
-function soft_drop(jet::PseudoJet, clusterseq::ClusterSequence,
-                   tag::SoftDropTagger)
-    rad = tag.cluster_rad
-    new_clusterseq = recluster(jet, clusterseq; R = rad, algorithm = JetAlgorithm.CA)
+function soft_drop(jet::PseudoJet, clusterseq::ClusterSequence; zcut::Float64,
+                   beta::Float64, radius = 1.0)
```
graeme-a-stewart, line 108: `zcut`, `beta` and `radius` should be subtypes of `Real`

src/Substructure.jl
```diff
@@ -204,18 +144,18 @@ Filters a jet to retain only the hardest subjets based on a specified radius and
...
+- `hardest_jets::Int64`: Number of hardest jets to retain in the filtered result.
 
 # Returns:
 - `PseudoJet`: Filtered jet composed of the hardest subjets.
 """
-function jet_filtering(jet::PseudoJet, clusterseq::ClusterSequence, filter::JetFilter)
-    rad = filter.filter_radius
-    new_clusterseq = recluster(jet, clusterseq; R = rad, algorithm = JetAlgorithm.CA)
+function jet_filtering(jet::PseudoJet, clusterseq::ClusterSequence; radius::Float64,
+                       hardest_jets::Int64)
```
graeme-a-stewart, line 154: `radius` should be `::Real`, `hardest_jets` should be `::Integer`

src/Substructure.jl
```diff
@@ -231,17 +171,18 @@ Trims a jet by removing subjets with transverse momentum below a specified fract
...
 
 # Returns:
 - `PseudoJet`: Trimmed jet composed of retained subjets.
 """
-function jet_trimming(jet::PseudoJet, clusterseq::ClusterSequence, trim::JetTrim)
-    rad = trim.trim_radius
-    alg = trim.recluster_method
-    frac2 = trim.trim_fraction^2
+function jet_trimming(jet::PseudoJet, clusterseq::ClusterSequence; radius::Float64,
+                      fraction::Float64, recluster_method::JetAlgorithm.Algorithm)
```
graeme-a-stewart, line 182: `radius` and `fraction` are `::Real`

grasph:
Concerning performance, you should also read https://docs.julialang.org/en/v1/manual/faq/#faq-nothing: last sentence of 3rd paragraph "Note that the compiler is able to generate efficient code when working with Union{T, Nothing} arguments or fields." 

That said, I like your approach. It has the advantage, that often no special treatment will be needed, as the null four-vector will be rejected by the analysis cut on pt or mass. `invalidPseudoJet` might be a better name, you can then have a isValid(::PseudoJet) function to embed the '===' trick. `invalidPseudoJet` is more generic and can used in more cases.

graeme-a-stewart:
> Concerning performance, you should also read https://docs.julialang.org/en/v1/manual/faq/#faq-nothing: last sentence of 3rd paragraph "Note that the compiler is able to generate efficient code when working with Union{T, Nothing} arguments or fields."

👍 

> 
> That said, I like your approach. It has the advantage, that often no special treatment will be needed, as the null four-vector will be rejected by the analysis cut on pt or mass. `invalidPseudoJet` might be a better name, you can then have a isValid(::PseudoJet) function to embed the '===' trick. `invalidPseudoJet` is more generic and can used in more cases.

I agree - let's use `invalid`. A lower case name and method would be more consistent with Julia conventions (just like the instance of the type `Nothing` is `nothing`; test for nothing is `isnothing()`), https://docs.julialang.org/en/v1/manual/style-guide/#Use-naming-conventions-consistent-with-Julia-base/.

- Invalid instance: `invalid_pseudojet`
- Test for validity: `isvalid()`

src/Pseudojet.jl
```diff
@@ -91,6 +91,25 @@ A PseudoJet object.
...
+# Arguments
+- `j::PseudoJet`: The `PseudoJet` object to check.
+
+# Returns
+- `Bool`: `true` if the `PseudoJet` object is non-zero, `false` otherwise. 
+"""
+function isvalid(j::PseudoJet)
+    return !(j.px == invalid_pseudojet.px && j.py == invalid_pseudojet.py &&
+             j.pz == invalid_pseudojet.pz && j.E == invalid_pseudojet.E)
+end
```
m-fila, line 111:
```suggestion
isvalid(j::PseudoJet) = !(j === invalid_pseudojet)
```

I think the suggestion was to do identity comparison (Is this an invalid_jet object?) rather than equality (Do they have the same properties?). The docstring might be also updated to better reflect this

m-fila (commented):
I think my preference would be to go with `nothing` as, I think, having a `const` object of `mutable struct` is quite error prone way to model  "invalid state". `const` means the `invalid_pseudojet` can be changed to another object but the current object can be still modified since it's a mutable struct.
If by mistake the `invalid_jet` property is modified (and it's very easy to do in user code) then the benefits of `invalid_jet` passing as a `PseudoJet`becomes possible problems. 
```julia
some_jet = invalid_pseudojet
some_jet.E= 3.14 # ups
invalid_pseudojet.E # 3.14 😱
```
In this example, if in any further calls a user gets a jet that might be invalid (calling `mass_drop` for instance) and then forgets to check it with `isvalid` because it will be filtered out later anyway, then they might be up for a surprise

Should `PseudoJet` be immutable then I wouldn't see any problems with having`invalid_pseudojet`

test/test-substructure.jl
```diff
@@ -16,35 +16,37 @@ for m in keys(methods)
...
             @test length(groomed) === length(test_data[ievt]["jets"])
             for (ijet, jet) in enumerate(groomed)
-                @test JetReconstruction.rapidity(jet)≈test_data[ievt]["jets"][ijet]["rap"] atol=tol
-                @test JetReconstruction.phi(jet)≈test_data[ievt]["jets"][ijet]["phi"] atol=tol
-                @test JetReconstruction.pt(jet)≈test_data[ievt]["jets"][ijet]["pt"] atol=tol
+                test_jet = JetReconstruction.isvalid(jet) ? jet :
+                           JetReconstruction.invalid_pseudojet
+                @test JetReconstruction.rapidity(test_jet)≈test_data[ievt]["jets"][ijet]["rap"] atol=tol
+                @test JetReconstruction.phi(test_jet)≈test_data[ievt]["jets"][ijet]["phi"] atol=tol
+                @test JetReconstruction.pt(test_jet)≈test_data[ievt]["jets"][ijet]["pt"] atol=tol
```
m-fila, line 49:
```suggestion
                @test JetReconstruction.rapidity(jet)≈test_data[ievt]["jets"][ijet]["rap"] atol=tol
                @test JetReconstruction.phi(jet)≈test_data[ievt]["jets"][ijet]["phi"] atol=tol
                @test JetReconstruction.pt(jet)≈test_data[ievt]["jets"][ijet]["pt"] atol=tol
```

checking if jet is valid then returning it or `invalid_pseudojet`  - this isn't needed anymore since if jet is invalid then is already should be an `invalid_pseudojet`

graeme-a-stewart:
> I think my preference would be to go with `nothing` as, I think, having a `const` object of `mutable struct` is quite error prone way to model "invalid state". `const` means the `invalid_pseudojet` can be changed to another object but the current object can be still modified since it's a mutable struct. If by mistake the `invalid_jet` property is modified (and it's very easy to do in user code) then the benefits of `invalid_jet` passing as a `PseudoJet`becomes possible problems.
> 
> ```julia
> some_jet = invalid_pseudojet
> some_jet.E= 3.14 # ups
> invalid_pseudojet.E # 3.14 😱
> ```
> 
> In this example, if in any further calls a user gets a jet that might be invalid (calling `mass_drop` for instance) and then forgets to check it with `isvalid` because it will be filtered out later anyway, then they might be up for a surprise
> 
> Should `PseudoJet` be immutable then I wouldn't see any problems with having`invalid_pseudojet`. Also see #145

As discussed on #145 we should try to make PseudoJet and EEJet immutable - there are quite a few benefits.

Then this issue of a user being able to accidentally update the `invalid_pseudojet` fields goes away.

src/Pseudojet.jl
```diff
@@ -91,6 +91,22 @@ A PseudoJet object.
 PseudoJet(px::Real, py::Real,
 pz::Real, E::Real) = PseudoJet(px, py, pz, E, 0, px^2 + py^2)
 
+"""Used to mark a jet incase the corresponding substructure tagging fails."""
```
graeme-a-stewart, line 94:
```suggestion
"""Used to mark an invalid result in case the corresponding substructure tagging fails."""
```

src/Pseudojet.jl
```diff
@@ -91,6 +91,22 @@ A PseudoJet object.
...
+const invalid_pseudojet = PseudoJet(0.0, 0.0, 0.0, 0.0)
+
+"""
```
graeme-a-stewart, line 97:
```suggestion

import Base.isvalid
"""
```

graeme-a-stewart: We should add this method to `Base.isvalid` to avoid namespace clashes

graeme-a-stewart (changes requested):
Hello @sattwamo - looks pretty good. I left a few small suggestions for the documentation and to add the `invalid` to `Base`.

graeme-a-stewart (approved):

PR #142, opened 2025-03-25
By default use local version of JetReconstruction in examples and docs

m-fila:
One too many times I forgot to `dev` the main package and was running with package from JuliaHub and wondering where are my changes

I propose to use [`[sources]`](https://pkgdocs.julialang.org/v1.11/toml-files/#The-[sources]-section) directive so examples docs automatically pick-up local version of JetReconstruction.jl. It's still possible to `free` or `dev` somewhere else later.  This also shouldn't  interfere with CI.

The `[sources]` directive was added in Julia 1.11. For earlier versions this is transparent and not doing anything, so the package should still be `dev`ed manually. 

Probably in 1.12 some of these could be replaced with [`[workspace]`](https://pkgdocs.julialang.org/dev/toml-files/#The-[workspace]-section) to resolve all the dependencies for sub-projects together.

### DIFF
--- a/docs/Project.toml
+++ b/docs/Project.toml
@@ -6,6 +6,9 @@ JetReconstruction = "44e8cb2c-dfab-4825-9c70-d4808a591196"
 LiveServer = "16fef848-5104-11e9-1b77-fb7a48bbb589"
 Makie = "ee78f7c6-11fb-53f2-987a-cfe4a2b5a57a"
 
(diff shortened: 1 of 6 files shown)


graeme-a-stewart (approved):
Hi @m-fila - I tested this with 1.9 just to check and indeed, it works just fine. Thanks!

PR #144, opened 2025-03-27
Jet utilities (#141)

graeme-a-stewart:
Add functions to calculate the momentum fraction and the kt scale for jets.

- pt_fraction() for transverse momentum fraction
- kt_scale()

These are implemented generically for both jet types, though may be primary interesting for pp events (PseudoJet).

deltar is added as a helper that returns the jet separation in pseudorapidity-phi. This is distinct from deltaR which operates in rapidity-phi space. These are probably not the greatest names, but are not in the public API. deltaR is moved to the new source file, JetUtils.jl.

There is a small tidy up of the logic for calculating ϕ where the check for ϕ>2π can never be true.

Added functions to convert jets to LorentzVector and LorentzVectorCyl. This required some initial refactoring of jet methods to be general, rather than PseudoJet specific, which will be continued.

(cherry picked from commit c7cfa9e300ed4741b903a88d936208519f9e2ea2)


graeme-a-stewart:
Checked on the 0.4 branch, so can be merged without further review

PR #146, opened 2025-04-11
Make jet structs immutable and improve jet struct methods

### LINKED ISSUE #143: Avoid duplicated methods for `EEJet` and `PseudoJet`
Currently a lot of methods are implemented in exactly the same way for `EEJet` and `PseudoJet`, which is unnecessary maintenance load.

These should be consolidated to

```Julia
some_method(jet::T) where {T <: FourMomentum}
```

With #133 the `JetUtils.jl` is introduced which is where these reimplementations should live.

### LINKED ISSUE #145: Make `PseudoJet` and `EEJet` isbits structs
Coming from #88, passing `mutable struct`s between C and Julia interfaces requires extra copies and allocations since they are differently stored in memory. This happen very often as the API consists of multiple smaller function passing around the `ClusterSequences` and jet objects. Unfortunately I don't have a concrete estimate on a cost of it.

Changing the `PseudoJet` and `EEJet` to be immutable `struct`s and isbits would allow one to remove that extra overhead.
I think it should also have a potential to improve slightly the performance on Julia side since then more continuous blocks of memory could be used.
Hopefully this should be possible with a moderate effort since the only moment when mutation is required seems to be adding history information

### LINKED ISSUE #157: Properly support alternative recombination schemes
To correctly support different recombination schemes there are actually two points that need to be addressed:

- Preprocessing input clusters to match the assumptions of the scheme (e.g., massless)
- Recombining two jets to form a new merged jet

At the moment only the second is supported in the reconstruction interfaces (via the `recombine` argument), so in fact different recombination schemes are **not** currently supported.

Note that #146 adds the complication that new merged jet needs to have it's cluster history value set correctly at construction time.

A proposed interface is to support passing *named* arguments, `preprocess` and `recombine`, which in practice can be done from a splatted named tuple.

graeme-a-stewart:
This PR makes the jet structures used in the package *immutable*. This is relatively straight forward, but some care is required with the `_cluster_hist_index`, which is the link from the jet to the correct point in the reconstruction history. This now has to be set correctly at the time of object construction. This is addressed in a few places:

- For constructing an initial vector of jets, it's necessary to pre-set the cluster history to `1...N`, so the initial jets are correctly lined up
  - This is done in the reconstruction entry points automatically if conversion to the internal EDM is happening
  - When reading from HepMC3 files to a `FourMomentum` struct, this is done

During the reconstruction it is easy to calculate the correct point in the history. This requires the `recombine` function to be restructured to have the signature `addjets(jet1, jet2; cluster_hist_index=index)`, which is not the same a simple `+`. However, it's not terrible and this is added to the documentation. Support is added for known schemes, currently massless Pt and Pt^2.

The bells-and-whistles reconstruction example, `instrumented-jetreco.jl`, can now take an `--recombine` option to use non-standard recombination.

Note that all of the jet struct constructors have been reworked to provide a more consistent interface and `cluster_hist_index` is always a *named* parameter. This PR also restructures the accessor/auxiliary methods for all the jet types, making these as generic as possible. To this end a new source file is added, `CommonJetStructs.jl` that factorises all of the generic type definitions and methods. Constructing jet types from LorentzVectors and LorentzVectorHEPs is made more consistent.

Core reconstruction times are slightly improved with these static structs, especially on x86_64.

For substructure functions, it was necessary to redo the `recluster()` method, to make *copies* of the relevant jets, with updated cluster indexes for the reclustering. This is actually a *bug fix* as the old way was updating the cluster indexes for the reclustering, which would break the reconstruction history for the original reconstruction. 

Docstrings were improved - adding missing ones and simplifying a bit others.

Closes #145 
Closes #143 
Closes #157


graeme-a-stewart:
CoPilot result - failure to launch!

graeme-a-stewart:
Just doing some performance testing, currently this is about the same speed as the current head of `main` (6a3e3a13e7f6727423b50dcd7c248cb0132a4c90).

The use of `PlainJet` is causing a performance loss, which is about the same as the gain from immutability. If this is coded out then there is a performance gain of ~4% for $pp$ reconstruction with the tiled strategy.

So avoiding this temporary is quite important (thus WIP until then).

graeme-a-stewart:
Another comment on performance. On my x86_64 benchmarking machine (AMD Ryzen 5700) the performance changes are:

| Reconstruction Algorithm | Speed up |
|---|---|
| Anti-Kt `N2Tiled` | +11% |
| Anti-Kt `N2Plain` | -0.3% |
| Durham | +7% |

This is a sweet little speed boost!

I scanned the `N2Plain` code, but it's nowhere obvious why this strategy is somehow loosing ground with the immutable structs.

m-fila (commented):
Since recombination and preprocessing are connected would it make sense to make them a single argument in the public facing API? I'm thinking about something like `recombination_scheme::NamedTuple{(:recombination, :preprocess)}` in signatures and pre-defined schemes
``` julia
const EScheme = (recombine = addjets_escheme, preprocess = nothing)
const PtScheme = (recombine = addjets_ptscheme, preprocess = preprocess_ptscheme)
const Pt2Scheme = (recombine = addjets_pt2scheme, preprocess = preprocess_pt2scheme))
```
I think it's slightly simpler than having the dictionary and enum, and if someone wants a custom scheme they just put in a named tuple and use the same mechanism.
On the other hand having an enum would be handy for C-bindings :slightly_smiling_face: 

---

There are a few mentions of  recombination with `+` in README and `src/TiledAlgoLL.jl`
- https://github.com/JuliaHEP/JetReconstruction.jl/blob/e7edcef43fb5c8710a07cfc9b95cb946fd374816/README.md#L29-L31
  And explanation of arguments below the function is missing `preprocess`
- https://github.com/JuliaHEP/JetReconstruction.jl/blob/e7edcef43fb5c8710a07cfc9b95cb946fd374816/README.md#L91-L94
- https://github.com/JuliaHEP/JetReconstruction.jl/blob/e7edcef43fb5c8710a07cfc9b95cb946fd374816/src/TiledAlgoLL.jl#L362-L365
- https://github.com/JuliaHEP/JetReconstruction.jl/blob/e7edcef43fb5c8710a07cfc9b95cb946fd374816/src/TiledAlgoLL.jl#L426-L435

graeme-a-stewart:
> Since recombination and preprocessing are connected would it make sense to make them a single argument in the public facing API? I'm thinking about something like `recombination_scheme::NamedTuple{(:recombination, :preprocess)}` in signatures and pre-defined schemes
> 
> ```julia
> const EScheme = (recombine = addjets_escheme, preprocess = nothing)
> const PtScheme = (recombine = addjets_ptscheme, preprocess = preprocess_ptscheme)
> const Pt2Scheme = (recombine = addjets_pt2scheme, preprocess = preprocess_pt2scheme))
> ```
> 
> I think it's slightly simpler than having the dictionary and enum, and if someone wants a custom scheme they just put in a named tuple and use the same mechanism. On the other hand having an enum would be handy for C-bindings 🙂

OTOH, the dictionary and enum is our business. And the two arguments, which can be passed from a splatted named tuple is pretty standard stuff. I did add a test in the CI that all of the members of the enum have an entry in the `RecombinationMethods` dictionary, that should help catch any desynchronising between the two structures.

graeme-a-stewart:
OK, I think this is finally all ready to go! I have just done a last set of performance measures on x86_64 (AMD Ryzen 5700) and on MacBook Pro M2. Performance diff with `main` is:

| Test | x86_84 | M2 Pro |
|---|---|---|
| N2Tiled | +11% | +4% |
| N2 Plain | +2% | -0.6% |
| Durham | +7% | 0% |

So still seeing a nice speed up for the most important platform and most important use cases.

graeme-a-stewart:
I have updated the initial commit comment to reflect the issues which were worked on in the process of this PR (working recombination schemes and improved jet structure methods).

m-fila (approved):
LGTM!

test/test-jet-types.jl
```diff
@@ -2,8 +2,8 @@
 
 include("common.jl")
 
-pj = PseudoJet(1.0, 2.0, 3.0, 10.0)
-pj._cluster_hist_index = 7
+pj = PseudoJet(1.0, 2.0, 3.0, 10.0; cluster_hist_index = 7)
+pj_beam = PseudoJet(0.0, 0.0, 5.0, 5.0)
 @testset "PseudoJet tests" begin
     @test JetReconstruction.px(pj) ≈ 1.0
```
m-fila, line 8:
```suggestion
@testset "PseudoJet tests" begin
    @test isbits(pj)
    @test JetReconstruction.px(pj) ≈ 1.0
```

test/test-jet-types.jl
```diff
@@ -14,9 +14,19 @@ pj._cluster_hist_index = 7
...
+    # This isn't really a test of the output, but rather that the object
+    # can be printed without error
+    @test string(pj) == "PseudoJet(px: 1.0 py: 2.0 pz: 3.0 E: 10.0 cluster_hist_index: 7)"
 end
 
-eej = EEJet(1.0, 2.0, 3.0, 10.0, 7)
+eej = EEJet(1.0, 2.0, 3.0, 10.0; cluster_hist_index = 7)
+eej_beam = EEJet(0.0, 0.0, 5.0, 5.0)
 @testset "EEJet tests" begin
     @test JetReconstruction.px(eej) ≈ 1.0
```
m-fila, line 31:
```suggestion
@testset "EEJet tests" begin
    @test isbits(eej)
    @test JetReconstruction.px(eej) ≈ 1.0
```

m-fila (commented):
Maybe one last thing. Since this was the origin of this development, could we add an explicit test that the jet types are `isbits`?

test/test-jet-types.jl
graeme-a-stewart: I like this!

PR #149, opened 2025-05-01
Update input particles to be AbstractVector

### LINKED ISSUE #147: example/EDM4hep not working
Currently, in the [**JetReconstruction.jl**/examples/EDM4hep](https://github.com/JuliaHEP/JetReconstruction.jl/tree/main/examples/EDM4hep), the two examples [EDM4hepJets.jl](https://github.com/JuliaHEP/JetReconstruction.jl/blob/main/examples/EDM4hep/EDM4hepJets.jl) and [SimpleRecoEDM4hep.jl](https://github.com/JuliaHEP/JetReconstruction.jl/blob/main/examples/EDM4hep/SimpleRecoEDM4hep.jl) are not working correctly. 

It is due to the reason that the `recps = RootIO.get(reader, evt, "ReconstructedParticles")` get from the EDM4hep.RootIO is returning `StructArrays.StructVector{ReconstructedParticle, @xxxxx` while we should expect the input to be `Vector{T}` in the `jet_reconstruct()` function.  

A quick fix I tried is `vector_recps = collect(recps)` and do `cs = jet_reconstruct(vector_recps; algorithm = JetAlgorithm.Durham).` There might possibly be more formal way to fix this.

graeme-a-stewart:
Take an `AbstractVector` for the algorithm entry points instead of a `Vector`, allowing any vector-like collection to be used.

In particular this allows EDM4hep data to be used correctly again.

This PR targets the current 0.4 release branch.

Closes #147

### DIFF
--- a/src/EEAlgorithm.jl
+++ b/src/EEAlgorithm.jl
@@ -258,13 +258,13 @@ function ee_genkt_algorithm(particles::AbstractVector{T}; p = 1,
 end
 
 """
-    _ee_genkt_algorithm(; particles::Vector{EEjet}, p = 1, R = 4.0,
+    _ee_genkt_algorithm(; particles::AbstractVector{EEjet}, p = 1, R = 4.0,
                        algorithm::JetAlgorithm.Algorithm = JetAlgorithm.Durham,
                        recombine = +)
 
 This function is the actual implementation of the e+e- jet clustering algorithm.
 """
-function _ee_genkt_algorithm(; particles::Vector{EEjet}, p = 1, R = 4.0,
+function _ee_genkt_algorithm(; particles::AbstractVector{EEjet}, p = 1, R = 4.0,
                              algorithm::JetAlgorithm.Algorithm = JetAlgorithm.Durham,
                              recombine = +)
     # Bounds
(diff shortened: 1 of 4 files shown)


m-fila (commented):
I didn't check it, but will it work also for `AbstractVector{PseudoJets}` which aren't `Vector``?
Taking plain jet as example:
- if `T` is `PseudoJet` make copy of input
- if `T` isn't `PseudoJet` make `Vector{PseudoJet}` and convert input to it
- call ` _plain_jet_reconstruct` which accepts only `Vector{PseudoJet}`

This seems like a problem when parameter is `PseudoJet` but type isn't `Vector`, no?

Perhaps we could add a test using StructArrays and pseudojets just to be sure

graeme-a-stewart:
Hi @m-fila - that's a very good point. I think you identified a failure mode I had not thought of. I have now widened the algorithm implementations to also take an `AbstractVector`, as well as put that type specification onto the main user entry point, `jet_reconstruct()`, so that if the user passes an unworkable type of particles then this gets rejected at the earliest stage possible.

m-fila (approved):
LGTM

Issue #83, opened 2024-10-25
Test removal of LoopVectorization.jl


Moelf:
Since LV.jl is going away in 1.12, maybe we should attempt to remove it.

I'm looking at https://github.com/graeme-a-stewart/JetReconstructionBenchmarks.jl/ and want to find a small set of benchmarks I can run to monitor the performance when making changes, @graeme-a-stewart can you help to point out a minimal set of things to run, using that repo? It's not completely clear what to run (e.g. example snippets)

Moelf:
I guess something like this:

```julia
akako@frame16 ~/D/g/JetReconstructionBenchmarks.jl (main)> julia --project src/benchmark.jl data/events-pp-13TeV-20GeV.hepmc3.gz
1×5 DataFrame
 Row │ File_path                          File                             mean_particles  n_samples  time_per_event
     │ Any                                String                           Int64           Int64      Float64
─────┼───────────────────────────────────────────────────────────────────────────────────────────────────────────────
   1 │ data/events-pp-13TeV-20GeV.hepmc…  events-pp-13TeV-20GeV.hepmc3.gz              -1         16         37.1709
akako@frame16 ~/D/g/JetReconstructionBenchmarks.jl (main)> julia --project src/benchmark.jl --backend Julia data/events-pp-13TeV-20GeV.hepmc3
1×5 DataFrame
 Row │ File_path                          File                          mean_particles  n_samples  time_per_event
     │ Any                                String                        Int64           Int64      Float64
─────┼────────────────────────────────────────────────────────────────────────────────────────────────────────────
   1 │ data/events-pp-13TeV-20GeV.hepmc3  events-pp-13TeV-20GeV.hepmc3              -1         16         44.7281
akako@frame16 ~/D/g/JetReconstructionBenchmarks.jl (main)> julia --project src/benchmark.jl --backend Julia data/events-pp-13TeV-20GeV.hepmc3
1×5 DataFrame
 Row │ File_path                          File                          mean_particles  n_samples  time_per_event
     │ Any                                String                        Int64           Int64      Float64
─────┼────────────────────────────────────────────────────────────────────────────────────────────────────────────
   1 │ data/events-pp-13TeV-20GeV.hepmc3  events-pp-13TeV-20GeV.hepmc3              -1         16         35.1953
```

but it seems quite unstable

Moelf:
on my machine, without `@turbo` is faster...

```
# with @turbo
[...]

PR #151, opened 2025-05-14
use SIMD.jl for x86 and `naive_findmin` for `:aarch64`

m-fila:
Just a test if I understand how to replace `fast_findmin` implementation as in #83


src/Utils.jl
```diff
@@ -147,13 +147,7 @@ array. The use of `@turbo` macro gives a significant performance boost.
...
-    # findmin(@inbounds @view dij[1:n])
-    best = 1
-    @inbounds dij_min = dij[1]
-    @turbo for here in 2:n
-        newmin = dij[here] < dij_min
-        best = newmin ? here : best
-        dij_min = newmin ? dij[here] : dij_min
-    end
-    dij_min, best
+    x = @fastmath foldl(min, dij)
```
Moelf, line 150: so I think the error is telling us that:
1. there are duplicated value in the distance vector
2. and this new algorithm is picking a different element to return than the current behavior

Interestingly, removing `fastmath` doesn't change the result, and it doesn't help if we change `findfirst` -> `findlast`

I'm a bit lost, what was LoopVectorization doing?

src/Utils.jl
```diff
@@ -147,13 +147,7 @@ array. The use of `@turbo` macro gives a significant performance boost.
...
+    i = findfirst(==(x), dij)::Int
+    x, i
```
Moelf, line 152:
```suggestion
    x = @fastmath foldl(min, @view(dij[begin:n]))
    i = findfirst(==(x), dij)::Int
    x, i
```

Moelf:
after fixing that, on Apple M4, I get:

```
# main
Processed 100 events 32 times
Average time per event 135.5111340625 ± 2.9835167009602035 μs
Lowest time per event 131.43084000000002 μs

# this PR
Processed 100 events 32 times
Average time per event 138.62045531249998 ± 3.5853189437051123 μs
Lowest time per event 134.35666 μs
```

this looks not bad??!? @graeme-a-stewart

src/Utils.jl
m-fila: thanks!

Moelf:
>AMD Ryzen 7 5700G

yeah ok so we use the SIMD.jl for x86 and this PR for ARM?

Moelf:
what happened to CI lol...

m-fila:
For completeness, with SIMD for x86_64

AMD Ryzen 7 5700G
```
# main
Processed 100 events 32 times
Average time per event 187.38033500000006 ± 5.74337091484025 μs
Lowest time per event 177.03909 μs

# this PR
Processed 100 events 32 times
Average time per event 191.52783343750005 ± 6.355636113548961 μs
Lowest time per event 179.90162 μs

# main with 1.12-beta3
Processed 100 events 32 times
Average time per event 195.5927890625 ± 12.11654841840912 μs
Lowest time per event 178.41858 μs

# this PR with 1.12-beta3
Processed 100 events 32 times
Average time per event 194.36534874999998 ± 10.480610090304218 μs
Lowest time per event 176.65239000000003 μs
```

Moelf:
https://github.com/JuliaLang/julia/pull/58418

I will test against this PR later

giordano:
Would be good to run CI on aarch64 (Linux and/or macOS), no?

Moelf:
```
julia #58418 on M4
Processed 100 events 32 times
Average time per event 151.89809874999995 ± 5.095528318872066 μs
Lowest time per event 143.37917 μs
```

so this is same as nightly, and slower than Julia 1.11

graeme-a-stewart:
I took a look at performance on my M2 Pro (where the original patch in #83 was sucky). For 1.11.4  I got:

```sh
 ~/.julia/dev/JetReconstruction/examples/ [main] ./benchmark.sh
pp 14TeV Tiled
Processed 100 events 16 times
Average time per event 171.11460874999997 ± 4.616430628080113 μs
Lowest time per event 165.90792000000002 μs
Processed 100 events 16 times
Average time per event 674.727474375 ± 18.35030922586858 μs
Lowest time per event 656.97167 μs
ee H Durham
Processed 100 events 16 times
Average time per event 28.442970625 ± 0.30894069468529917 μs
Lowest time per event 27.61167 μs
```

and with this patch:

```sh
 ~/.julia/dev/JetReconstruction/examples/ [naive_findmin] ./benchmark.sh 256
pp 14TeV Tiled
Processed 100 events 16 times
Average time per event 175.676691875 ± 12.098880482560652 μs
Lowest time per event 167.775 μs
pp 14 TeV Plain
Processed 100 events 16 times
Average time per event 671.201953125 ± 5.289937381395405 μs
Lowest time per event 662.80833 μs
ee H Durham
Processed 100 events 16 times
Average time per event 28.853904375 ± 0.5296152717583557 μs
Lowest time per event 27.69208 μs
```

Performance loss is marginal - we can live with it.

I also looked at 1.12.0-beta3:

 ~/.julia/dev/JetReconstruction/examples/ [main] ./benchmark.sh
pp 14TeV Tiled
Processed 100 events 16 times
Average time per event 186.92546749999997 ± 10.934245123593346 μs
Lowest time per event 175.50416 μs
pp 14 TeV Plain
Processed 100 events 16 times
Average time per event 684.5363275000001 ± 11.514369132124589 μs
Lowest time per event 674.45708 μs
ee H Durham
Processed 100 events 16 times
Average time per event 29.18406375 ± 1.3979762370840427 μs
Lowest time per event 28.45625 μs
```

```sh
 ~/.julia/dev/JetReconstruction/examples/ [naive_findmin] ./benchmark.sh
pp 14TeV Tiled
Processed 100 events 16 times
Average time per event 189.20445375 ± 8.512180852002729 μs
Lowest time per event 177.52208 μs
pp 14 TeV Plain
Processed 100 events 16 times
Average time per event 683.996173125 ± 9.641498982006677 μs
Lowest time per event 674.17833 μs
ee H Durham
Processed 100 events 16 times
Average time per event 30.5012775 ± 0.44837712099670635 μs
Lowest time per event 29.78125 μs
```

So the performance loss with 1.12 is more significant and we should report that.

Moelf:
>So the performance loss with 1.12 is more significant and we should report that.

I can't think of anything specific that would cuase this. In fact, I think the micro benchmark for `findmin()` shows that 1.12 and nightly are better than 1.11.5, so I suspect there may be regression in somewhere else.

Another supporting evidence is that even with LV.jl, 1.12 is slower, and LV.jl should be independent of Julia versions (I don't believe LLVM regressed )

graeme-a-stewart:
FYI I also want to merge the immutable structs patch, #146,  before considering merging this one

m-fila:
Just to note,  LoopVectorization seems to be broken with Julia nightly for 2 weeks or so

m-fila:
There was another problem in context of static compilation since LoopVectorization prevents trimming and also is super slow with PackageCompiler

PR #152, opened 2025-05-15
run ci with arm

m-fila:
Add running CI on GitHub hosted Arm Linux runners

https://github.com/JuliaHEP/JetReconstruction.jl/pull/151#issuecomment-2883399269

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -12,25 +12,26 @@ concurrency:
   cancel-in-progress: ${{ startsWith(github.ref, 'refs/pull/') }}
 jobs:
   test:
-    name: Julia ${{ matrix.version }} - ${{ matrix.os }} - ${{ matrix.arch }} - ${{ github.event_name }}
-    runs-on: ${{ matrix.os }}
+    name: Julia ${{ matrix.version }} - ${{ matrix.platform.os }} - ${{ matrix.platform.arch }} - ${{ github.event_name }}
+    runs-on: ${{ matrix.platform.os }}
     strategy:
       fail-fast: false
       matrix:
+        platform:
+          - os: ubuntu-latest
+            arch: x64
+          - os: ubuntu-24.04-arm
+            arch: aarch64
         version:
           - '1.9'
           - '1'
           - 'nightly'
-        os:
-          - ubuntu-latest
-        arch:
-          - x64
     steps:
       - uses: actions/checkout@v4
       - uses: julia-actions/setup-julia@v2
         with:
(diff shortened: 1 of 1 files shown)


Moelf (approved):

.github/workflows/CI.yml
```diff
@@ -12,25 +12,26 @@ concurrency:
...
-    name: Julia ${{ matrix.version }} - ${{ matrix.os }} - ${{ matrix.arch }} - ${{ github.event_name }}
-    runs-on: ${{ matrix.os }}
+    name: Julia ${{ matrix.version }} - ${{ matrix.platform.os }} - ${{ matrix.platform.arch }} - ${{ github.event_name }}
+    runs-on: ${{ matrix.platform.os }}
     strategy:
       fail-fast: false
       matrix:
+        platform:
+          - os: ubuntu-latest
+            arch: x64
```
giordano, line 22: Why setting `arch` at all?

.github/workflows/CI.yml
```diff
@@ -12,25 +12,26 @@ concurrency:
...
+          - os: ubuntu-24.04-arm
```
graeme-a-stewart, line 23: Shouldn't we use `ubuntu-latest` here as well? It is odd to have an inconsistency between x86 and ARM.


More broadly, I think `ubuntu-latest` is sufficient here. Julia has far fewer distro specific issues than something that might have to compile with the native c++ compiler.

graeme-a-stewart (changes requested):

.github/workflows/CI.yml
giordano: No.  `ubuntu-latest` is x86-64, `ubuntu-24.04-arm` is the aarch64 image, there's no "latest" alias, see https://github.com/actions/partner-runner-images/

graeme-a-stewart: OK - I didn't realise that, thanks! Looking at https://github.com/marketplace/actions/run-on-architecture#supported-platforms, is it then `ubuntu_latest` for ARM, to achieve the same result? My mild concern is that we end up leaving the ARM distro behind the x86 one.

graeme-a-stewart (approved):

.github/workflows/CI.yml
giordano: I think that's using qemu, which you really don't want to use if you can possibly avoid it.

graeme-a-stewart: OK, we go with this PR as written then!

PR #155, opened 2025-05-19
use LorentzVectorHEP and not LorentzVectors

ExpandingMan:
Currently this package depends on *both* `LorentzVectors` and `LorentzVectorHEP`, even though the latter is a re-implementation of the former.  This removes `LorentzVectors` entirely in favor of `LorentzVectorHEP` only.


ExpandingMan:
Also, tests are currently on 1.9, latest and nightly but *not* 1.10, which is the current LTS.  Would you be willing to drop 1.9 tests but include 1.10 tests?

Moelf (approved):
I have no problem with this except.

idk what will happen when we fully use LorentzVectorBase but I imagine it won't be breaking for LorentzVectorHEP.jl

ExpandingMan:
This can always get changed as needed, it just seemed to me that the situation as of right this very moment is that
- LorentzVectorBase is not registered and not used by LorentzVectorHEP
- This package depends on methods form LorentzVectorHEP
- LorentzVectorHEP completely (or nearly completely?) reimplements what's in LorentzVectors without depending on it.

So it seemed to me that the dual dependencies did not make sense.

.github/workflows/CI.yml
```diff
@@ -23,7 +23,7 @@ jobs:
           - os: ubuntu-24.04-arm
             arch: aarch64
         version:
-          - '1.9'
+          - '1.10'
```
m-fila, line 26: I think that was to test the minimal version supported in Project.toml

graeme-a-stewart: Yeah, this is a better idea - then we are testing LTS, the latest v1 and the nightly 👍

graeme-a-stewart:
I'd like to start to work on using LorentzVectorBase. For now this is a nice clean up and I like the idea of testing on LTS, latest and nightly. Thanks @ExpandingMan!

graeme-a-stewart (approved):

graeme-a-stewart:
@all-contributors please add @ExpandingMan for infra

PR #161, opened 2025-05-26
Make `algorithm` argument mandatory in `jet_reconstruct`

### LINKED ISSUE #129: Deprecate calling jet_reconstruct with only a power
Currently, for historical reasons, we support calling `jet_reconstruct()` specifying only the value for `p`, where $p \in -1,0,1$, which we then map to $pp$ reconstruction as `AntiKt`, `CA` or `Kt` (it dates from the 0.3 series, when only $pp$ was available).

However, it's an *implicit* selection of the algorithm, which is clear than being explicit.

Therefore, we should change the interface to make `algorithm` a **mandatory** argument. `p` remains optional and only has to be given in those cases where the algorithm needs it (currently `EEKt` and `GenKt`).

N.B. This can be done by changing `algorithm` to a positional parameter,

```Julia
function jet_reconstruct(particles, algorithm::Union{JetAlgorithm.Algorithm, Nothing};
                         p::Union{Real, Nothing} = nothing, R = 1.0,
                         recombine = +,
                         strategy = RecoStrategy.Best)
```

or by keeping it named, but making it mandatory,

```Julia
function jet_reconstruct(particles; algorithm::Union{JetAlgorithm.Algorithm, Nothing},
                         p::Union{Real, Nothing} = nothing, R = 1.0,
                         recombine = +,
                         strategy = RecoStrategy.Best)
```

IMO keeping it a named parameter is more consistent with the current interface and also results in clearer user code.

See also #128.

m-fila:
This is a **breaking change**

The `algorithm` keyword argument will be mandatory in `jet_reconstruct`, which effectively removes calling `jet_reconstruct` with `power`. The power will be still a keyword argument:
- required for generalized algorithms (GenKt and EEKt)
- ignore by other algorithms (appropriate power will be assigned internally instead of using the value from argument)

This allows to simplify the logic a bit and also will be easier to support in C-bindings #88
A few internal functions are removed. The internal `algorithm2power` dictionary now maps both pp and ee algorithms to their power, dictionary going in opposite direction is removed

Internal algorithms (`_plain_jet_reconstruct` etc) assume that `algorithm` and `p` given to them is consistent and won't check it or resolve again

Closes #129


README.md
```diff
@@ -27,7 +27,7 @@ algorithm and generalised $`k_\text{T}`$ for $`e^+e^-`$.
 The simplest interface is to call:
 
 ```julia
-cs = jet_reconstruct(particles::AbstractVector{T}; algorithm = JetAlgorithm.AntiKt, R = 1.0, [p = -1,] [strategy = RecoStrategy.Best])
+cs = jet_reconstruct(particles::AbstractVector{T}; algorithm = JetAlgorithm.AntiKt, R = 1.0, [p = nothing,] [strategy = RecoStrategy.Best])
```
graeme-a-stewart, line 30: I would suggest we don't even mention `p` (or `strategy`) here. This is the simplest get-up-and-go level of the documentation and it should give the minimum requirement for running.

README.md
```diff
@@ -42,21 +42,11 @@ cs = jet_reconstruct(particles::AbstractVector{T}; algorithm = JetAlgorithm.Anti
   - `JetAlgorithm.Durham` the $e^+e-$ $k_\text{T}$ algorithm, also known as the Durham algorithm
   - `JetAlgorithm.EEKt` the $e^+e-$ generalised $k_\text{T}$ algorithm
 - `R` - the cone size parameter; no particles more geometrically distance than `R` will be merged (default 1.0; note this parameter is ignored for the Durham algorithm)
+- `p` - algorithm power which must be provided for generalised algorithms, for other algorithms this parameter is ignored
```
graeme-a-stewart, line 45: Drop `p` and `strategy` descriptions.

We could link to https://juliahep.github.io/JetReconstruction.jl/stable/#Reconstruction-Interface as a "see here for additional options"

graeme-a-stewart (changes requested):
Thank you for doing this @m-fila! It looks very complete to me, just a few minor issues to clean up in the `README.md`.

graeme-a-stewart (approved):
I think this is GTG now!

PR #171, opened 2025-06-26
Remove jet serialisation to ASCII

### LINKED ISSUE #169: Deprecate custom serialisation functions
In the early days of this package functionality was added to serialise and deserialise jets (as `LorentzVectorHEP`) from an ASCII file.

This is unused and should now be removed (`Serialise.jl`).

m-fila:
Removing unused jet serialisation to ASCII
I assumed we want to skip the deprecation and just remove it since it's supposed to be part of 1.0 release

Closes #169


graeme-a-stewart:
Yes, I don't think there's any need to go through a deprecation process here

Issue #172, opened 2025-06-27
PseudoJet history index assertion failing in dummy example


m-fila:
I observe assertion errors in a dummy example I was using to populate precompilation statements:

```
using JetReconstruction

R = 2.0
algorithm = JetAlgorithm.AntiKt
strategy = RecoStrategy.N2Plain

particles = [PseudoJet(0.0, 1.0, 2.0, 3.0), PseudoJet(1.0, 2.0, 3.0,
                                                      4.0)]

jet_reconstruct(particles, algorithm = algorithm, R = R, strategy = strategy)
```

Before changing PseudoJet to immutable it worked without issues. Since #146 I get:
```
ERROR: AssertionError: cluster_hist_index(particles[i]) == i
Stacktrace:
 [1] initial_history(particles::Vector{PseudoJet})
   @ JetReconstruction ~/JetReconstruction2/src/ClusterSequence.jl:95
 [2] _plain_jet_reconstruct(particles::Vector{PseudoJet}; algorithm::JetReconstruction.JetAlgorithm.Algorithm, p::Int64, R::Float64, recombine::typeof(addjets))
   @ JetReconstruction ~/JetReconstruction2/src/PlainAlgo.jl:321
 [3] plain_jet_reconstruct(particles::Vector{PseudoJet}; algorithm::JetReconstruction.JetAlgorithm.Algorithm, p::Nothing, R::Float64, recombine::Function, preprocess::Nothing)
   @ JetReconstruction ~/JetReconstruction2/src/PlainAlgo.jl:263
 [4] plain_jet_reconstruct
   @ ~/JetReconstruction2/src/PlainAlgo.jl:225 [inlined]
 [5] jet_reconstruct(particles::Vector{…}; algorithm::JetReconstruction.JetAlgorithm.Algorithm, p::Nothing, R::Float64, recombine::Function, preprocess::Nothing, strategy::JetReconstruction.RecoStrategy.Strategy)
   @ JetReconstruction ~/JetReconstruction2/src/GenericAlgo.jl:85
 [6] top-level scope
   @ REPL[6]:1
```

I probably should switch to get pre-compilation statements from the tests which are more realistic (and runs fine), but this might indicate there is a gap in the logic

PR #174, opened 2025-07-02
Document requirements for input jets' history indices

m-fila:
I propose to add a short notice that the jets used as input for `jet_reconstruct` must have their cluster history indices properly set. I got confused by this in #172.
Should a similar notice appear also for `plain_jet_reconstruct`, `tiled_jet_reconstruct`, `ee_genkt_algorithm`?

Also fix some more minor typos I found on the way

### DIFF
--- a/src/ClusterSequence.jl
+++ b/src/ClusterSequence.jl
@@ -92,7 +92,7 @@ function initial_history(particles)
 
         # get cross-referencing right from the Jets
         # particles[i]._cluster_hist_index = i
-        @assert cluster_hist_index(particles[i]) == i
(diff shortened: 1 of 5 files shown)


Moelf:
this looks like there should be a wrapper type or constructor to set the index right automatically

graeme-a-stewart (changes requested):
Thanks! One minor change suggested.

m-fila:
Thanks, indeed now the message is less confusing. I also added message to the assert

graeme-a-stewart (approved):

PR #175, opened 2025-07-04
Fix missing algorithm arg in examples in docstrings

m-fila:
This a follow-up to #161. Some of the examples in docstrings were missing the `algorithm` argument


Moelf:
should we make these doctests?

m-fila:
I'm not sure, these functions are for specific strategies, the main one that people use and read about was updated. I think removing these examples is also an option (there are no examples for the EE variant)

PR #179, opened 2025-07-11
Make `SoftKiller` immutable

m-fila:
Follow-up to #153 making the `SoftKiller` an immutable struct. Also reorder the fields so min and max rapidity are in the same order in the constructor and struct

@emadmtr 

Also changing `Int64` to `Int` for consistency

PR #180, opened 2025-07-11
Use concrete int type for fields in HepMC3 `Particle` struct

m-fila:
Spotted when checking integer types during #179. The `Particle` struct in HepMC3 had fields of abstract type, but was only ever used with `Int`.


graeme-a-stewart:
It was excluded from benchmarks, which we only ever did once the change to the internal EDM (PseudoJet, EEJet) had been done. It was terribly suboptimal for sure...

PR #182, opened 2025-07-16
fix dump in jetreco and instrumented-jetreco examples

### LINKED ISSUE #181: Intended use of examples/jetreco.jl dump option?
Hi all,

I am working with a summer student, trying to use the `examples/jetreco.jl` `--dump` option to write some of the jets made by the example out for further study. 

It looks like activating the option attempts to push the jets back into a `jet_collection` array or list that doesn't exist:

https://github.com/mattleblanc/JetReconstruction.jl/blob/main/examples/jetreco.jl#L72

Should this just be added to the example somewhere, or are we missing something about the intended way to run and save output from this script?

Thanks a lot!
Matt

m-fila:
The `--dump` option for `jetreco.jl` and `instrumented-jetreco.jl` was broken because either variable declaration was missing or type conversion was missing

Closes #181 

@mattleblanc

### DIFF
--- a/examples/.gitignore
+++ b/examples/.gitignore
@@ -6,3 +6,5 @@
 profile/*
 # Playground for development scripts and notebooks
 playground/*
+# Ignore outputs from tests
+test-output*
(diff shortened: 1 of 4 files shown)


Moelf:
can this be tested? (run examples?)

PR #187, opened 2025-07-17
Fix type instability when passing type to a function, concretize default types

### LINKED ISSUE #185: Type instability in `inclusive_jets` and `exclusive_jets`
While experimenting with `--trim=safe` in #88 I noticed we have an instability in `inclusive_jets` and `exclusive_jets`. Both functions take a type of jets to use for the return as a keyword argument `T`

https://github.com/JuliaHEP/JetReconstruction.jl/blob/24631147985e1eb11a454f07a86ada653f54b50d/src/ClusterSequence.jl#L248-L249

One could think that it's harmless since yeah  the return type can be inferred from `T`, but nope, keyword arguments [don't participate](https://docs.julialang.org/en/v1/manual/methods/#Note-on-Optional-and-keyword-Arguments) in dispatch hence the type isn't static and we get instability.

(Couriusly adding annotation for `T::Type{V}` and then using `V` doesn't work as it's still a keyword argument issue but silences a few errors for `juliac --trim=safe`)

- `@code_warntype inclusive_jets(cs; ptmin = ptmin, T=LorentzVectorCyl)` this is unstable
  ```
  MethodInstance for Core.kwcall(::@NamedTuple{ptmin::Float64, T::UnionAll}, ::typeof(inclusive_jets), ::ClusterSequence{PseudoJet})
    from kwcall(::NamedTuple, ::typeof(inclusive_jets), clusterseq::ClusterSequence{U}) where U @ JetReconstruction ~/JetReconstruction/src/ClusterSequence.jl:248
  Static Parameters
    U = PseudoJet
  Arguments
    _::Core.Const(Core.kwcall)
    @_2::@NamedTuple{ptmin::Float64, T::UnionAll}
    @_3::Core.Const(JetReconstruction.inclusive_jets)
    clusterseq::ClusterSequence{PseudoJet}
  Locals
    ptmin::Union{}
    T::Union{}
    @_7::Union{Float64, UnionAll}
  Body::Vector
  ```

- `@code_warntype inclusive_jets(cs; ptmin = ptmin)` to make things even spicier this on the other hand is stable
  ```
  MethodInstance for Core.kwcall(::@NamedTuple{ptmin::Float64}, ::typeof(inclusive_jets), ::ClusterSequence{PseudoJet})
    from kwcall(::NamedTuple, ::typeof(inclusive_jets), clusterseq::ClusterSequence{U}) where U @ JetReconstruction ~/JetReconstruction/src/ClusterSequence.jl:248
  Static Parameters
    U = PseudoJet
  Arguments
    _::Core.Const(Core.kwcall)
    @_2::@NamedTuple{ptmin::Float64}
    @_3::Core.Const(JetReconstruction.inclusive_jets)
    clusterseq::ClusterSequence{PseudoJet}
  Locals
    ptmin::Union{}
    T::Union{}
    @_7::Union{Float64, Type{LorentzVectorCyl}}
  Body::Vector{LorentzVectorCyl}
  ```

I'm not sure how we want to handle this.
- Adding type as a first positional argument seems to be common practice (`convert` etc) but might be cumbersome here since type shenanigan aren't really the main spotlight our functions
- Perhaps we could return the same type as `ClusterSequence` and then separately if needed users then can convert to whatever they want?
- Or maybe we want to always return `LorentzVectorCyl`? For sake of #88 I could not provide `T` and use the default, and say that from C-bindings you only get `LorentzVectorCyl` (and for now only use `PseudoJet` as input) but that's just avoiding the real problem

@Moelf @graeme-a-stewart

m-fila:
As described in #185, passing a type as a keyword argument to a function causes a type instability. Passing type as a positional argument is fine. [reference](https://docs.julialang.org/en/v1/manual/methods/#Note-on-Optional-and-keyword-Arguments)

I propose here to change keyword argument `T=DefaultT` in `inclusive_jets`, `exclusive_jets`, etc. to be the last positional argument (and also constrain it to be `Type`), This eliminates the type instability but changes the interface. The last positional argument is a different convention that `Base.convert` and other standard functions but they don't allow the default value which we want to have

Another thing as pointed by @Moelf  is that parametric types used by in the default values (or used in examples) could be concretized.

Performance wise I don't see any noticeable difference with `julia --project instrumented-jetreco.jl --algorithm=AntiKt -R 0.4 ../test/data/events.pp13TeV.hepmc3.zst -m 32` due to the type instability

- main https://github.com/JuliaHEP/JetReconstruction.jl/blob/24631147985e1eb11a454f07a86ada653f54b50d/examples/instrumented-jetreco.jl#L166
  ```
  Processed 100 events 32 times
  Average time per event 128.2739265625 ± 3.2664981607124073 μs
  Lowest time per event 122.77002 μs
  ```
- main instrumented-jetreco.jl#L166 changed to `inclusive_jets(cs; ptmin = ptmin, T=LorentzVectorCyl)` (should be the worst)
  ```
  Processed 100 events 32 times
  Average time per event 128.9642315625 ± 3.379545577028563 μs
  Lowest time per event 122.93741 μs
  ```
- this PR with instrumented-jetreco.jl#L166 changed to `inclusive_jets(cs, LorentzVectoCyl; ptmin = ptmin)`
  ```
  Processed 100 events 32 times
  Average time per event 127.5241915625 ± 2.8969402308373184 μs
  Lowest time per event 122.15337 μs
  ```

- this PR with instrumented-jetreco.jl#L166 changed to `inclusive_jets(cs, LorentzVectorCyl{Float64}; ptmin = ptmin)`
  ```
  Processed 100 events 32 times
  Average time per event 126.56505062500004 ± 3.578129031677597 μs
  Lowest time per event 120.3919 μs
  ```

Also `preprocess_ptscheme` had wrong name in docstring and wasn't using `jet_type` argument

Fixes #185
Closes #186

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,7 +53,7 @@ To obtain the final inclusive jets, use the `inclusive_jets` method:
 final_jets = inclusive_jets(cs::ClusterSequence; ptmin=0.0)
 ```
 
-Only jets passing the cut $p_T > p_{Tmin}$ will be returned. The result is returned as a `Vector{LorentzVectorHEP}`, but different return types can be specified (e.g., `T = EEJet`).
+Only jets passing the cut $p_T > p_{Tmin}$ will be returned. The result is returned as a `Vector{LorentzVectorHEP}`, but different return types can be specified (e.g., `inclusive_jets(cs::ClusterSequence, EEJet; ptmin=0.0)`).
 
 #### Sorting
 
--- a/docs/src/softkiller.md
+++ b/docs/src/softkiller.md
@@ -33,7 +33,7 @@ A typical workflow using SoftKiller is:
 using JetReconstruction
 
 # Read event particles (see examples/softkiller/softkiller_plots.jl)
-particles = read_final_state_particles("event.hepmc3", T=PseudoJet)[1]
+particles = read_final_state_particles("event.hepmc3", PseudoJet)[1]
 
 # Set up SoftKiller
 sk = SoftKiller(rapmax=5.0, grid_size=0.4)
--- a/examples/EDM4hep/EDM4hepJets.jl
+++ b/examples/EDM4hep/EDM4hepJets.jl
@@ -56,7 +56,7 @@ function main()
         if args[:printjets]
             @info begin
                 jets = "Event $(ievt)\n"
-                for jet in exclusive_jets(cs; njets = args[:njets], T = EEJet)
+                for jet in exclusive_jets(cs, EEJet; njets = args[:njets])
                     jets *= " $jet\n"
                 end
                 jets
--- a/examples/EDM4hep/SimpleRecoEDM4hep.jl
+++ b/examples/EDM4hep/SimpleRecoEDM4hep.jl
@@ -14,7 +14,7 @@ recps = RootIO.get(reader, evt, "ReconstructedParticles")
 
 # Reconstruct and print the jets
 cs = jet_reconstruct(recps; algorithm = JetAlgorithm.Durham)
-dijets = exclusive_jets(cs; njets = 2, T = EEJet)
+dijets = exclusive_jets(cs, EEJet; njets = 2)
 for jet in dijets
     println(jet)
 end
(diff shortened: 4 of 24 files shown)


graeme-a-stewart (approved):

graeme-a-stewart:
This is great! Thanks @m-fila. I had missed the issue with _named_ arguments back when I added the type option to the jet selection functions. It's very useful that you picked this up with the pre-compilation work on the package.

PR #190, opened 2025-07-26
Only plot histogram when possible

graeme-a-stewart:
Do not try to plot a histogram with less than 2 values (it crashes!).

Discovered when I ran over a HepMC3 file with just one event in it.


graeme-a-stewart:
Trivial change to a utility - self-merging

PR #193, opened 2025-07-27
Improved softkiller examples

graeme-a-stewart:
Rename `--maxevents` and `--skip` to `--pileup-maxevents` and `--pileup-skip` which allow variations in the amount of pileup to be applied.

Select the hard scatter event with `--eventno` (this only ever one event).

Change hard scatter and pileup files to be positional arguments, which is more consistent with other scripts.

Make both scripts a bit more verbose for the user's information.

Run a jet reconstruction in `softkiller_runtime.jl` so that we actually do something with the reduced event.

N.B. for now, the reduced event has to be rewritten into a new vector to get the `cluster_hist_index`es correct. This can be removed once softkiller returns a good-for-reconstruction output.

Add a `test.sh` script that can be used for a quick test of the examples.

### DIFF
--- a/examples/softkiller/softkiller_runtime.jl
+++ b/examples/softkiller/softkiller_runtime.jl
@@ -11,15 +11,20 @@ include(joinpath(@__DIR__, "..", "parse-options.jl"))
 function parse_command_line(args)
     s = ArgParseSettings(autofix_names = true)
     @add_arg_table! s begin
-        "--maxevents", "-n"
-        help = "Maximum number of events to read. -1 to read all events from the  file."
+        "--pileup-maxevents", "-n"
+        help = "Maximum number of pileup events to read. -1 to read all events from the pileup events file."
         arg_type = Int
         default = -1
 
-        "--skip", "-s"
-        help = "Number of events to skip at beginning of the file."
+        "--pileup-skip", "-s"
+        help = "Number of events to skip in the pileup events file."
         arg_type = Int
         default = 0
+
+        "--eventno", "-e"
+        help = "Event number to process from the hard scatter events file. If not specified, the first event will be processed."
+        arg_type = Int
+        default = 1
 
         "--ptmin"
         help = "Minimum p_t for final inclusive jets (energy unit is the same as the input clusters, usually GeV)"
@@ -52,20 +57,17 @@ function parse_command_line(args)
         arg_type = RecoStrategy.Strategy
         default = RecoStrategy.Best
 
-        "--dump"
-        help = "Write list of reconstructed jets to a JSON formatted file"
-
         "--grid-size"
         help = "Size of Rectangular grid"
         arg_type = Float64
         default = 0.4
 
-        "--hard-file"
-        help = "HepMC3 event file in HepMC3 to read."
+        "hard-file"
+        help = "HepMC3 event file containing hard scatter events."
         required = true
 
-        "--pileup-file"
-        help = "HepMC3 event file in HepMC3 to read."
+        "pileup-file"
+        help = "HepMC3 event file containing pileup events."
         required = true
     end
     return parse_args(args, s; as_symbols = true)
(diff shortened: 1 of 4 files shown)


emadmtr:
Hi @graeme-a-stewart, I checked the changes everything seems great! I only have a question for some of the arguments that we can pass from the command line, because I haven't implemented a numer of them anywhere so I was just curious if you have a plan for them or they could be removed.

graeme-a-stewart:
If you go to the "Files Changed" tab there's a "Review Changes" button and you can approve this PR if you think everything is ok now.

emadmtr (approved):

examples/softkiller/softkiller_runtime.jl
```diff
@@ -110,18 +99,30 @@ function main()
...
-    end
+    # Fill hard event jets from first read event in h_events (actually should be only one event!)
+    append!(all_jets_sk, h_events[1])
 
     # Apply SoftKiller to all_jets_sk (hard + pileup)
-    softkiller(soft_killer, all_jets_sk)
+    reduced_event, pt_threshold = softkiller(soft_killer, all_jets_sk)
+    @info "SoftKiller applied: $(length(reduced_event)) clusters remaining from $(length(all_jets_sk)), pt threshold = $pt_threshold"
+
+    # Workaround for cluster_hist_indedx not being correct after SoftKiller filtering
```
mattleblanc, line 112: Maybe not for this PR, but I wanted to say that I find this procedure to be very cumbersome, in-part because it has to happen between applying filters to the input constituents and any jet finding that is done later.

SK isn't the only filter people will want to study: CS and CHS are also obvious things to look at, and grooming jets will have a similar problem. 

Is there some smarter way to handle the cluster_hist_index assignment when filtering particles? In FastJet I guess that this is part of the `ClusterSequence` and so it's handled there instead of by the user ...

mattleblanc (approved):

examples/softkiller/softkiller_runtime.jl
m-fila: Both in Fastjet and JetReconstruction the cluster history indices are part of the jet classes. The problem here is that our `PseudoJet`s are Julia's struct and are immutable so their fields (index) can't be updated later so we have to create a new jet object with different index. Prior to https://github.com/JuliaHEP/JetReconstruction.jl/pull/146, we had `mutable struct` which can be modified but had a number of other annoyances (reasoning about code with constants , less performant, more awkward C-bindings)

I guess as a minimum we could add a function taking a vector of jets after filter and returning vector of jets with normalized indices or maybe doing this inplace

graeme-a-stewart: Hello @mattleblanc 

Your point is very well taken. We started to put effort into ensuring that this would be correct, but there are many cases to cover and lots of different paths to failure!

I propose that the reconstruction algorithms themselves take responsibility for checking and fixing things if a collection is passed that has mis-set the `cluster_hist_index`. This should not cost much at runtime and it makes things much safer and prevents the code from bombing if there's a filtering/selection path that didn't fix-up the `cluster_hist_index` itself.

See #200.

AFAICR Fastjet is using mutable structures, which is inherently more dangerous for the code, but allows these kins of fix-ups on the fly. But we have ways to make sure that this is correct in Julia as well.

mattleblanc: I think that having the clustering algorithms themselves handle this makes the most sense, even if it costs a little bit. In this case, at least we can ensure it's done as optimally as possible, instead of leaving the job up to the user.

Mutable structs seem undesirable for the reasons you both mention...

Moelf: we  could consider writing custom https://juliaobjects.github.io/Accessors.jl/dev/examples/custom_optics/ if it makes everyone's life easier while keeping strict immutable

PR #194, opened 2025-07-27
Run examples in CI

### LINKED ISSUE #95: Test examples in CI
The examples are important as guidance for using the package. As things evolve it is really important to keep them working, so they should be tested in the CI.

To keep things lean, this also needs #94.

graeme-a-stewart:
The PR adds support for running examples in the CI, using the extended environment available for testing.

This should be suitable for running tests with lightweight additional dependencies. Graphical things, which need Makie, are still too heavy IMO.

The old example test scripts (which didn't work) are removed.

Fixes #95.


graeme-a-stewart:
Paths seem to be not quite correct yet

Moelf:
>Basic jetreco.jl reconstruction examples |    3      3  24.7s

seems to be a bit slow -- do we always want to run these, or reduce the # of events, or make them opt in like https://github.com/JuliaConcurrent/Atomix.jl/blob/771d9fecca321551576518f0eb3ace2358d4887e/test/runtests.jl#L138-L139

Moelf (approved):

Moelf:
>If it's too slow for local testing 

I think one could argue 30s may be fine

PR #195, opened 2025-07-27
Internal function improvements

### LINKED ISSUE #192: Avoid unnecessary copies of particles in "inner" jet reconstruction (and a few small internal interface fixes)
In the `_tiled_jet_reconstruct()` function (the one that does the actual jet reconstruction), there is a copy taken of the input particles:

https://github.com/JuliaHEP/JetReconstruction.jl/blob/1270f05dedda571f466b835de58fe445891b73ae/src/TiledAlgoLL.jl#L410-L417

This is unnecessary as the user facing `tiled_jet_reconstruct()` already makes a copy for the algorithm's internal use.

We should avoid the extra copy and change the name of `_tiled_jet_reconstruct()` to `_tiled_jet_reconstruct!()` just to make it clear that we will mutate the particles vector that gets passed in there.

This is not the case for `_plain_jet_reconstruct()` although there is a misleading comment indicating that a copy is made:

https://github.com/JuliaHEP/JetReconstruction.jl/blob/1270f05dedda571f466b835de58fe445891b73ae/src/PlainAlgo.jl#L322-L326

For `_ee_genkt_algorithm()` there are no copies made either, however there is an oddity that `particles` is a named argument and it should be positional, to be consistent with the pp algorithms.

Also, `_plain_jet_reconstruct` -> `_plain_jet_reconstruct!` and `_ee_genkt_algorithm` -> `_ee_genkt_algorithm!` for the same reasons as above.

graeme-a-stewart:
Prevent unnecessary copying of particle data for tiled jet reconstruction (this gains ~1% in runtime 😁).

Rename all internal reconstruction functions with a `!` as they mutate the particle data passed to them (it seeds the cluster sequence objects).

Make particles a positional argument for `_ee_genkt_algorithm!` to be consistent with pp versions.

Improve docstrings.

Fixes #192.

### DIFF
--- a/src/EEAlgorithm.jl
+++ b/src/EEAlgorithm.jl
@@ -263,8 +263,8 @@ function ee_genkt_algorithm(particles::AbstractVector{T}; algorithm::JetAlgorith
 
     # Now call the actual reconstruction method, tuned for our internal EDM
     _ee_genkt_algorithm!(recombination_particles; p = p, R = R,
-                        algorithm = algorithm,
-                        recombine = recombine)
+                         algorithm = algorithm,
+                         recombine = recombine)
 end
 
 """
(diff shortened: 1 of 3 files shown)


graeme-a-stewart:
Thanks for spotting the extra docstring issues @m-fila

PR #204, opened 2025-09-19
test ee reconstruction with EEJet type

m-fila:
The `run_reco_test` function which is used to test the correctness of jet reconstruction always starts from `PseudoJet` input jets.
While ee algortihms can be called with `PseudoJet` input (which will be later converted to EEJet) it probably make more sense to test it from `EEJet` input from the start


graeme-a-stewart (approved):

graeme-a-stewart:
Thanks @m-fila - a nice little optimisation!

PR #205, opened 2025-09-22
fix `Qtot` type in `ClusterSequence` dostrings

m-fila:
A small documentation issue. `ClusterSeuqence` filed `Qtot` is `Float64` yet the docstrings where referring to it as `Any`


src/ClusterSequence.jl
```diff
@@ -145,11 +145,11 @@ Construct a `ClusterSequence` object.
 - `jets::Vector{T}`: The jets in the cluster sequence, which are of T <: FourMomentum
 - `history::Vector{HistoryElement}`: The branching history of the cluster
   sequence.
-- `Qtot::Any`: The total energy of the event.
+- `Qtot::Float64`: The total energy of the event.
 """
 function ClusterSequence(algorithm::JetAlgorithm.Algorithm, p::Real, R::Float64,
                          strategy::RecoStrategy.Strategy, jets::Vector{T}, history,
-                         Qtot) where {T <: FourMomentum}
+                         Qtot::Float64) where {T <: FourMomentum}
```
Moelf, line 152: Why does the constructor need to be restricted? Can we not get away with anything that can be converted to a `Float64`?

graeme-a-stewart: I agree Jerry - it was written in an overly specific way. I removed the unneeded types and the explicit conversion step.

The constructor is outer because it's a convenience method, rather than enforcing some invariant.

graeme-a-stewart (approved):

Issue #162, opened 2025-05-26
AirspeedVelocity.jl for benchmark in CI


Moelf:
https://discourse.julialang.org/t/easy-github-benchmarking-with-new-airspeedvelocity-jl/129327

I noticed we have a separate repo in https://github.com/graeme-a-stewart/JetReconstructionBenchmarks.jl/ , I guess we can checkout that repo during CI

@graeme-a-stewart do you think this is a good idea?

graeme-a-stewart:
This looks very interesting - we should investigate. JetReconstructionBenchmarks.jl is way too much. The value here is to quickly identify regressions in the main use cases that we have. If you look at `examples/benchmark.sh` those are the baseline measurements to take to show that none of the algorithms regressed.

m-fila:
Just to keep track, there are some more alternatives like [Tachometer.jl](https://github.com/KristofferC/Tachometer.jl) and [PerfChecker](https://mirage-interactive-fr.github.io/PerfChecker/dev/)

PR #209, opened 2025-11-08
add benchmarks compatible with PkgBenchmark

m-fila:
Adding benchmark suite in format expected by PkgBenchmark and other tools.
Also adding a runner to execute locally from command line. `--verbose` displays histogram and statistics for each case

This could be potentially later used by some CI #162


graeme-a-stewart:
Great, thanks @m-fila - this is indeed useful.

I added a patch to update the e+e- benchmarks:

- Durham algorithm takes no parameters (neither distance nor p)
- EEKt algorithm takes two parameters: distance *and* p

Issue #191, opened 2025-07-26
SoftKiller should return PseudoJets with the correct `cluster_hist_index`


graeme-a-stewart:
Currently SoftKiller is copying out `PseudoJet` objects *verbatim* using `filter`:

https://github.com/JuliaHEP/JetReconstruction.jl/blob/1270f05dedda571f466b835de58fe445891b73ae/src/SoftKiller.jl#L176

There are two issues here:

1. This is quite expensive, but at the moment I don't see a way to get around this: we need a real vector of clusters to then initiate jet reconstruction, which will be the next step in the chain (if we were to return, e.g., references, then we'd need to construct the clusters vector at a later stage anyway), so we're just paying a necessary cost up-front that needs to be paid.
2. The `cluster_hist_index` values are wrong, because they come from the cluster order as read from `read_final_state_particles`.

The consequence of 2 is that the user needs to recreate the `reduced_event` *again* with the correct sequential values for the `cluster_hist_index`. e.g. as done here: 
https://github.com/JuliaHEP/JetReconstruction.jl/blob/1270f05dedda571f466b835de58fe445891b73ae/examples/softkiller/softkiller_plots.jl#L88-L92).

This also means that the [example](https://github.com/JuliaHEP/JetReconstruction.jl/blob/1270f05dedda571f466b835de58fe445891b73ae/docs/src/softkiller.md?plain=1#L42-L45) given in the documentation is *wrong* as the reconstruction will crash when the `@assert` which checks for sequential `cluster_hist_index` will fail.

So we're paying the price of two copies, when we only really need one.

So, I propose that instead of using `filter` we loop over the event's clusters and create new clusters, with the correct linear ordering for `cluster_hist_index`.

N.B. for reasons of efficiency it is important to use the *fully specified constructor* for PseudoJet, e.g.,

```Julia
new_cluster = PseudoJet(c.px, c.py, cpz, c.E, the_new_cluster_hist_index, c. _pt2, c._inv_pt2, c._rap, c._phi)
```

This will avoid the expensive recalculation of the internally cached values.

emadmtr:
[...]

PR #211, opened 2025-11-16
Add E-scheme preprocessor fixing cluster history indices

### LINKED ISSUE #200: Automatically fix-up jet `cluster_hist_seq` when (re)running reconstruction
Since the internal jet types became immutable we put some effort into ensuring that vectors of `PseudoJet` and `EEJet` are *created* with the correct sequencing of `cluster_hist_seq` (e.g., when reading input clusters from HepMC3 files).

However, this is pretty onerous on the user, given that we also now provide more functionality for filtering particles and re-clustering.

The proposal is that at the entry point into the reconstruction algorithms we check that the correct `cluster_hist_seq` is set. If it's not, then we fix it automatically using `Accessors.jl`.

This should be a very lightweight check if all is well, but still quite fast if a fix-up is required.

This removes the burden on users and developers to make sure that this is always correct.

m-fila:
The `ptscheme` and `pt2scheme` preprocessors already assign correct history indices to jet objects. The default `escheme` does not currently have its own preprocessor: it either copies the full vector of jets when they are already `PseudoJet`/`EEJet` (assuming their cluster indices are correct), or it converts to `PseudoJet`/`EEJet` and assigns new history indices during that conversion.

As described in #200, relying on the input jets for `escheme` to already have correct cluster indices can be easy to overlook. Instead of performing checks and index fix-ups as suggested there, this proposal introduces a dedicated preprocessor for `escheme` that always copies jets and assigns correct history indices. This would become the default behavior for e-scheme (other scheme preprocessors already include fixing indices). Users who know their input jets already have valid history indices can opt out by specifying `nothing` as the preprocessor to avoid unnecessary work.

I also noticed that two forms of preprocessor functions exist: one that accepts an output type and performs conversion, and one that does not. The docs describes user-provided preprocessors as the non-converting type, but the implementation only uses the converting form. I fix the docs and remove the unused methods.

Finally, this PR includes fixes for history-index in SoftKiller (#191), as it is related and there has been no recent activity on that issue.

Closes #200 #191

### DIFF
--- a/test/test-jet-utils.jl
+++ b/test/test-jet-utils.jl
@@ -49,10 +49,10 @@ eej2 = EEJet(-1.0, 3.2, -1.2, 39.0)
     @test JetReconstruction.energy(pj1_alt) ≈ JetReconstruction.energy(pj1)
 
     pj1_ptyetaphi = PseudoJet(; pt = JetReconstruction.pt(pj1),
-                               rap = JetReconstruction.rapidity(pj1),
-                               phi = JetReconstruction.phi(pj1),
-                               m = JetReconstruction.mass(pj1),
-                               cluster_hist_index = 7)
+                              rap = JetReconstruction.rapidity(pj1),
+                              phi = JetReconstruction.phi(pj1),
+                              m = JetReconstruction.mass(pj1),
+                              cluster_hist_index = 7)
     @test JetReconstruction.px(pj1_ptyetaphi) ≈ JetReconstruction.px(pj1)
     @test JetReconstruction.py(pj1_ptyetaphi) ≈ JetReconstruction.py(pj1)
     @test JetReconstruction.pz(pj1_ptyetaphi) ≈ JetReconstruction.pz(pj1)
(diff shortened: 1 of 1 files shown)


m-fila:
Performance wise it seems calling the preprocess for fixing indices isn't more expensive than `nothing` preprocess just coping input jets without fixups:

- `preprocess_escheme`
  ```
  pp 14TeV Tiled
  Processed 100 events 16 times
  Average time per event 177.05440187500005 ± 18.650765441555805 μs
  Lowest time per event 161.72569 μs
  pp 14 TeV Plain
  Processed 100 events 16 times
  Average time per event 504.513385625 ± 18.07148126798454 μs
  Lowest time per event 486.39413 μs
  ee H Durham
  Processed 100 events 16 times
  Average time per event 32.36774750000001 ± 3.0916310772200886 μs
  Lowest time per event 29.67452 μs
  ```
- `nothing`
  ```
  pp 14TeV Tiled
  Processed 100 events 16 times
  Average time per event 176.36583312499997 ± 17.66222657424268 μs
  Lowest time per event 162.19935999999998 μs
  pp 14 TeV Plain
  Processed 100 events 16 times
  Average time per event 501.3887362499999 ± 16.77077525920156 μs
  Lowest time per event 485.47028 μs
  ee H Durham
  Processed 100 events 16 times
  Average time per event 35.764484375 ± 4.081166524662498 μs
  Lowest time per event 29.568640000000002 μs
  ```

PR #202, opened 2025-08-11
Valencia jets 🍊

mattleblanc:
It seemed cleanest to just open a new PR. The original description is below, I've also included the initial feedback from @Moelf.

---

This PR aims to implement the Valencia (VLC) jet clustering algorithm as another option to study in the context of lepton colliders. Many thanks to my summer student @EthanLynn916 for the initial julia implementation that this PR is based on!

The implementation follows the description in [1404.4294](https://arxiv.org/abs/1404.4294), and so depends on several parameters:

    R: The jet radius parameter.
    β: Corresponds to the existing power p of the algorithm.
    γ: The angular exponent parameter used in the Valencia beam distance.

The extra flexibility given by the VLC algorithm provides some handles that allow for additional suppression of backgrounds due to ISR / BIB / etc. that present themselves in high-energy lepton collisions. It was originally developed for use at CLIC, and is of interest to study in the context of a future muon collider.

This PR touches EEAlgorithm to add the necessary distance functions and parameters. I think this new code is factorized and does not change the performance of the existing algorithms, but am continuing to validate the work.

In particular, some of the new tests are not yet passing due to larger differences in jet rapidity than I'd expected. I'm hopeful that https://github.com/JuliaHEP/JetReconstruction.jl/pull/198 might help bring things slightly closer to the C++ output, so am waiting to test with that merged ... in the meantime, I'm happy to get feedback on this draft PR so that it can be improved.

🍻 MLB

PR #217, opened 2025-12-12
Valencia jets revisited 🍊

graeme-a-stewart:
This is a rebased and updated version of the Valencia jets implementation from @mattleblanc and @EthanLynn916.

Compared to the original in #202 there is:

- rebase to current head of main, incorporating changes to the default preprocessor from #211 
- some internal simplification (e.g., `invR2` is now mandatory at the *internal* `_ee_genkt_algorithm!` level, and β is handled at the user API, `ee_genkt_algorithm`)
- unnecessary checks in hot functions are removed to speed things up
- `--algorithm=Valencia` is now supported in the reconstruction examples `jetreco.jl` and `instrumented-jetreco.jl`; `--beta` is supported as an alias for `--power` for use with this algorithm

There is still a slight performance regression compared to the previous implantation of the algorithm when running Durham or EEKt, but this is now O(5%), which is considered acceptable. Options that improved even that small difference had too much of an impact on the sustainability of the code because they required a lot of code duplication.

Further internal optimisation could still be looked in to.

As noted in #202 this implementation follows the description in [1404.4294](https://arxiv.org/abs/1404.4294), and so depends on several parameters:

```
R: The jet radius parameter.
β: Corresponds to the existing power p of the algorithm.
γ: The angular exponent parameter used in the Valencia beam distance.
```

### DIFF
--- a/src/AlgorithmStrategyEnums.jl
+++ b/src/AlgorithmStrategyEnums.jl
@@ -37,7 +37,7 @@ const AllJetRecoAlgorithms = [String(Symbol(x)) for x in instances(JetAlgorithm.
 A constant array that contains the jet algorithms for which power is variable.
 
 """
-const varpower_algorithms = [JetAlgorithm.GenKt, JetAlgorithm.EEKt, JetAlgorithm.Valencia]
+const varpower_algorithms = (JetAlgorithm.GenKt, JetAlgorithm.EEKt, JetAlgorithm.Valencia)
 
 """
     algorithm2power
--- a/src/EEAlgorithm.jl
+++ b/src/EEAlgorithm.jl
@@ -32,7 +32,7 @@ Valencia metric (independent of `dij_factor`).
 @inline function dij_dist(eereco, i, j, dij_factor, algorithm, invR2)
     j == 0 && return large_dij
     @inbounds begin
-        if algorithm == JetAlgorithm.Valencia
+        if algorithm === JetAlgorithm.Valencia
             return valencia_distance(eereco, i, j, invR2)
         else
             # Durham & EEKt share same form here (min(E2p_i,E2p_j) * dij_factor * angular_metric)
@@ -97,7 +97,7 @@ function get_angular_nearest_neighbours!(eereco, algorithm, dij_factor, p, invR2
     N = length(eereco)
     # Initialise sentinels so the first comparison always wins
     @inbounds for i in 1:N
-        if algorithm == JetAlgorithm.Valencia
+        if algorithm === JetAlgorithm.Valencia
             eereco.nndist[i] = Inf
         else
             eereco.nndist[i] = large_distance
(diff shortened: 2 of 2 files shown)


examples/benchmark.sh
```diff
@@ -11,3 +11,6 @@ julia --project instrumented-jetreco.jl --algorithm=AntiKt -R 0.4 ../test/data/e
 
 echo "ee H Durham"
 julia --project instrumented-jetreco.jl --algorithm=Durham ../test/data/events.eeH.hepmc3.zst -m $trials
+
+echo "ee Valencia"
```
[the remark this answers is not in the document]
graeme-a-stewart: Yes, I agree - done now

src/AlgorithmStrategyEnums.jl
```diff
@@ -36,7 +37,7 @@ const AllJetRecoAlgorithms = [String(Symbol(x)) for x in instances(JetAlgorithm.
 A constant array that contains the jet algorithms for which power is variable.
 
 """
-const varpower_algorithms = [JetAlgorithm.GenKt, JetAlgorithm.EEKt]
+const varpower_algorithms = [JetAlgorithm.GenKt, JetAlgorithm.EEKt, JetAlgorithm.Valencia]
```
Moelf: should this be a tuple / static array?

src/EEAlgorithm.jl
```diff
@@ -24,113 +23,217 @@ Calculate the angular distance between two jets `i` and `j` using the formula
...
+end
+
+function get_angular_nearest_neighbours!(eereco, algorithm, dij_factor, p, invR2, γ = 1.0)
     # Get the initial nearest neighbours for each jet
     N = length(eereco)
-    # this_dist_vector = Vector{Float64}(undef, N)
-    # Nearest neighbour geometric distance
+    # Initialise sentinels so the first comparison always wins
+    @inbounds for i in 1:N
+        if algorithm == JetAlgorithm.Valencia
```
Moelf:
```suggestion
        if algorithm === JetAlgorithm.Valencia
```

src/EEAlgorithm.jl
```diff
@@ -24,113 +23,217 @@ Calculate the angular distance between two jets `i` and `j` using the formula
...
+    # Scan all other jets, update i, and cross-update j
     @inbounds for j in 1:N
         if j != i
-            this_nndist = angular_distance(eereco, i, j)
-            better_nndist_i = this_nndist < eereco[i].nndist
-            eereco.nndist[i] = better_nndist_i ? this_nndist : eereco.nndist[i]
-            eereco.nni[i] = better_nndist_i ? j : eereco.nni[i]
-            if this_nndist < eereco[j].nndist
-                eereco.nndist[j] = this_nndist
+            this_metric = algorithm == JetAlgorithm.Valencia ?
```
Moelf:
```suggestion
            this_metric = algorithm === JetAlgorithm.Valencia ?
```

src/EEAlgorithm.jl
```diff
@@ -24,113 +23,217 @@ Calculate the angular distance between two jets `i` and `j` using the formula
...
+            better_nndist_j = this_metric < eereco[j].nndist
+            eereco.nndist[j] = better_nndist_j ? this_metric : eereco.nndist[j]
             eereco.nni[j] = better_nndist_j ? i : eereco.nni[j]
         end
     end
     # Nearest neighbour dij distance
-    for i in 1:N
-        eereco.dijdist[i] = dij_dist(eereco, i, eereco[i].nni, dij_factor)
+    @inbounds for i in 1:N
+        if algorithm == JetAlgorithm.Valencia
```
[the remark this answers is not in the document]
graeme-a-stewart: I do not think so, from reading the lowered code.

PR #227, opened 2026-01-27
Access cached values when available

### LINKED ISSUE #224: Ensure cached jet values are used also for `LorentzVectorBase` methods
As noted in #223 `LorentzVectorBase.pt2(::PseudoJet)` will not use the cached value in this type.

To prevent this inefficiency from happening, appropriate methods for `::EEJet` and `::PseudoJet` should be added in the `LorentzVectorBase` module.

graeme-a-stewart:
For LorentzVectorBase accessors use cached values in PseudoJet and EEJet when these exist.

For `LorentzVectorBase.phi(::PseudoJet)` there is no use of the cached value because of the different range conventions in LorentzVectorBase, [-π, π], vs. JetReconstruction, which uses [0, 2π].

The generic accessors for `::FourVector` are not changed, as there is no guarantee that all concrete types would have any particular cached field (should more be FourVector types be introduced).

Closes #224

### DIFF
--- a/src/PseudoJet.jl
+++ b/src/PseudoJet.jl
@@ -211,4 +211,4 @@ Return the scalar transverse momentum (pt) of a PseudoJet. This accessor uses
 the precalculated value that the struct has.
 """
 pt(p::PseudoJet) = sqrt(p._pt2)
-LorentzVectorBase.pt(p::PseudoJet) = p._pt
+LorentzVectorBase.pt(p::PseudoJet) = sqrt(p._pt2)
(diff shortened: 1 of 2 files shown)


graeme-a-stewart:
Pretty simple patch, so self-merging this one

graeme-a-stewart:
Super! Thanks for having a look. Right now, I see no compelling reason for custom coordinates as these types represent particles that have well defined (energy, momentum) 4-vector properties.

PR #229, opened 2026-02-02
Add cluster_hist_index argument for EDM4hep reconstructed particles

### LINKED ISSUE #228: Bug for EEJet(EDM4hep.ReconstructionParticle)
Currently, in latest two versions of EEJet, when the input is EDM4hep.ReconstructedParticle, as the method in EDM4hepJet Extension doesn't have the cluster_hist_index, the method goes automatically to the function EEJet(jet::Any; cluster_hist_index::Int = 0) in src/EEJet.jl, line 113, and falls in error of line 120.

This occurs when perform this line: cs = jet_reconstruct(recps; p = 1.0, R = 2.0, algorithm = JetAlgorithm.EEKt)
where recps is the reconstructed particle from edm4hep event.

### LINKED ISSUE #230: Add a test for EDM4hep inputs
As found in #228 the lack of testing for EDM4hep inputs is hurting us.

We need to add a test that runs over a few events and checks that all is well for EE reconstruction from EDM4hep files.

This will meaning adding a few dependencies to the test environment,

graeme-a-stewart:
It is necessary for construction of EEJets from `::ReconstructedParticle` that the `cluster_hist_index` is an optional argument so that the calls from the jet preprocessor (setting the correct `cluster_hist_index`) work.

This has been checked against the examples in `examples/EDM4hep`, both of which now work.

A basic test case has been introduced that tests that EEJet and PseudoJet can be constructed from an `EDM4hep.edm4hep!ReconstructedParticle`. This requires adding EDM4hep to the `test` environment.

Closes #228 
Closes #230

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -48,9 +48,10 @@ julia = "1.10"
 [extras]
 Aqua = "4c88cf16-eb10-579e-8560-4a9242c79595"
 ArgParse = "c7e460c6-2fb9-53a9-8c5b-16f535851c63"
+EDM4hep = "eb32b910-dde9-4347-8fce-cd6be3498f0c"
 JSON = "682c06a0-de6a-54ab-a142-c8b1cf79cde6"
 Random = "9a3f8284-a2c9-5f02-9a11-845980a1fd5c"
 Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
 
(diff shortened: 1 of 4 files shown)


m-fila (approved):
Seems fine, checked with some example files 👍

PR #240, opened 2026-04-28
Implement active jet area calculation with explicit ghosts (fixes #124)

### LINKED ISSUE #124: Implement jet area calculations
The area of a jet has several applications when associating different objects using ghosts, and when performing calibrations, particularly for subtraction of the average contribution from pile-up to a jet:

https://arxiv.org/abs/2007.02645

https://arxiv.org/pdf/1107.4277

The methods used to calculate the jet are are documented by Cacciari et al., and available in `fastjet`:

https://arxiv.org/abs/0802.1188

https://github.com/alisw/fastjet/blob/a7a459d280f151197bc24676a243d9cafa065b81/fastjet/include/fastjet/ClusterSequenceAreaBase.hh

etc.

HarshitNagpal29:
Dear @graeme-a-stewart, @m-fila, @mattleblanc,

I have tried implementing an initial solution for #124. This PR adds a first version of active jet area calculation using explicit ghost particles, inspired by the FastJet implementation and written to fit the existing style of this package.

## What This Adds

- `GhostedAreaSpec` for configuring the ghost grid
- `ClusterSequenceArea` as a wrapper around `ClusterSequence`
- `jet_reconstruct_area(...)`
- Scalar jet area lookup with `area(jet, csa)`
- Area 4-vector lookup with `area_4vector(jet, csa)`
- Pure ghost identification with `is_pure_ghost(jet, csa)`
- `total_area` and ghost metadata helpers
- Tests for area propagation, pure-ghost filtering, default area-queryable jets, strategy rejection, and an anti-kt area sanity check

## Current Scope

This first implementation supports pp algorithms with explicit ghost active areas. It intentionally does not yet implement passive areas, Voronoi areas, rho estimation, or pile-up subtraction.

For now, `jet_reconstruct_area` only supports `RecoStrategy.N2Plain`.

## Note About N2Tiled

While adding tests for jet areas, I hit an existing issue with the `N2Tiled` strategy when used with ghosted areas:

```text
BoundsError: attempt to access 6×15 Matrix{TiledJet} at index [94]
```
My current understanding is that the tiling matrix is allocated as (n_tiles_eta, n_tiles_phi), but tile_index can allow iphi == n_tiles_phi, which can produce a linear index beyond the allocated matrix size. In my case the maximum valid index should be 90, but it reached 94.

Because of this, the area implementation currently rejects tiled strategies and uses only N2Plain.

## Testing

I added tests for the new area functionality. On my machine, the full test suite passes

## Docs

I have not updated the docs yet, because I thought it would be better to get feedback on the API and implementation approach first.

## AI usage

AI was used to help understand parts of the FastJet implementation and to generate some of the tests. I reviewed the generated material and take responsibility for the code and tests included in this PR.


Project.toml
```diff
@@ -1,7 +1,7 @@
 name = "JetReconstruction"
 uuid = "44e8cb2c-dfab-4825-9c70-d4808a591196"
-version = "1.0.0"
 authors = ["Graeme A Stewart <graeme.andrew.stewart@desy.de>", "Philippe Gras <philippe.gras@cern.ch>", "Atell Krasnopolski <delta_atell@protonmail.com>", "Mateusz Fila <mateusz.jakub.fila@cern.ch>"]
+version = "1.0.0"
```
HarshitNagpal29, line 4: also, should i bump the version?

m-fila: I'd say no, it will be easier to track if versions are bumped separately

HarshitNagpal29: ok, no problem

src/JetAreas.jl
```diff
@@ -0,0 +1,365 @@
...
+        else
+            is_pure_ghost[i] = false
+            areas[i] = 0.0
+            area_4vectors[i] = _zero_pseudojet(cluster_hist_index = i)
+        end
+    end
+
+    has_dangerous_particles = false
+    if max_ghost_pt2 > 0.0
+        danger_ratio = eps(Float64)^2
```
HarshitNagpal29, line 173: also, can you please guide on this its little bit confusing for me that is it fine or overaggressive?

src/JetAreas.jl
```diff
@@ -0,0 +1,335 @@
...
+struct GhostGridSpec
+    nphi::Int
+    nrap::Int
+    dphi::Float64
+    drap::Float64
+    actual_ghost_area::Float64
+    n_ghosts::Int
+end
+
+function _ghost_grid_spec(spec::GhostedAreaSpec)::GhostGridSpec
```
m-fila, line 34:
```suggestion
function _ghost_grid_spec(spec::GhostedAreaSpec)
```
Casting to `GhostGridSpec` isn't really needed here

Also, would it make sense to make this function a constructor of `GhostGridSpec`?

src/JetAreas.jl
```diff
@@ -0,0 +1,335 @@
...
+    return GhostGridSpec(nphi, nrap, dphi, drap, actual_ghost_area, n_ghosts)
+end
+
+function _copy_as_pseudojet(jet; cluster_hist_index::Int)
+    PseudoJet(px(jet),
+              py(jet),
+              pz(jet),
+              E(jet);
+              cluster_hist_index = cluster_hist_index,)
+end
```
m-fila, line 58: We have a generic constructor for `PseudoJet` https://github.com/JuliaHEP/JetReconstruction.jl/blob/3e96272eac5bb2dc77cca733c774b41f3d12bde0/src/PseudoJet.jl#L152
I think there is no need to introduce a new function, the `PseudoJet` already has a constructor accepting any jet

src/JetAreas.jl
```diff
@@ -0,0 +1,335 @@
...
+                              area_spec::GhostedAreaSpec = GhostedAreaSpec(),
+                              strategy = RecoStrategy.N2Plain,
+                              recombine = addjets_escheme,
+                              preprocess = preprocess_escheme,)
+    if algorithm ∉ (JetAlgorithm.Kt,
+        JetAlgorithm.CA,
+        JetAlgorithm.AntiKt,
+        JetAlgorithm.GenKt)
+        throw(ArgumentError("jet_reconstruct_area currently supports pp algorithms only"))
+    end
```
m-fila, line 233:
```suggestion
    if !is_pp(algorithm)
        throw(ArgumentError("jet_reconstruct_area currently supports pp algorithms only"))
    end
```
We have a function checking if algorithm is pp

src/JetAreas.jl
```diff
@@ -0,0 +1,335 @@
...
+                        ::Type{T} = PseudoJet;
+                        ptmin = 0.0,
+                        include_pure_ghosts::Bool = false,) where {T}
+    jets = inclusive_jets(csa.clusterseq, PseudoJet; ptmin = ptmin)
+
+    if !include_pure_ghosts
+        jets = filter(jet -> !is_pure_ghost(jet, csa), jets)
+    end
+
+    return [_convert_area_result_jet(jet, T) for jet in jets]
```
m-fila, line 334: I think it could would be nice to avoid allocating a new array if `T` is already `PseudoJet` and there is nothing to do

src/JetAreas.jl
```diff
@@ -0,0 +1,335 @@
...
+    return PseudoJet(scale * px(jet),
+                     scale * py(jet),
+                     scale * pz(jet),
+                     scale * E(jet);
+                     cluster_hist_index = cluster_hist_index,)
+end
+
+function _make_area_ghosts(spec::GhostedAreaSpec)
+    grid = _ghost_grid_spec(spec)
+    rng = isnothing(spec.seed) ? Random.default_rng() : Random.MersenneTwister(spec.seed)
```
m-fila, line 78: More Julia-idiomatic way would be to pass an `rng::AbstractRNG` to a function rather than create it inside

m-fila (commented):
A few comments from me, I didn't check the "algorithmic" part yet

PR #241, opened 2026-05-05
make jet index mismatch error message more descriptive

m-fila:
One too many times I got this error and was wondering what went wrong.
Adding information about expected and actual values of indices in the error message


graeme-a-stewart (approved):
Yes, this is clearly better - thanks!

PR #242, opened 2026-05-15
Add C-bindings and static compilation

m-fila:
This is revised version of #88 

Adds:
- module with C-bindings
- compilation with either `JuliaC.jl` or `PackageCompiler.jl`

The C-bindings for now cover only:
- `PseudoJet` and pp-reconstruction
- `exclusive_jets`, `inclusive_jets`

Updating from #88 and earlier versions of JetReconstruction revealed that updating the enums is quite error prone. To mitigate it, the enums in C-header are now configured directly from package source instead of being hardcoded


compile/downstream/.gitignore
```diff
@@ -0,0 +1 @@
+build/
```
graeme-a-stewart: Very minor - no newline

graeme-a-stewart (approved):
This is great @m-fila - thank you, I have not really any further suggestions on the code as is. It worked fine for me on Linux (RHEL 9.8).

However, on OS X I failed to compile, getting a really weird error from `juliac`:

```
juliac Compiling...LLVM ERROR: Cannot select: 0x8c095dbd0: i64 = vscale Constant:i64<1>, /Users/graemes/.julia/packages/HostCPUFeatures/ZTXz4/src/cpu_info_aarch64.jl:5 0x8c0964380: i64 = Constant<1> In function: julia_vscale_59980
```

Nasty and looks very low level ☹️. I have checked 1.13, still happens; I have checked the nightly, still happens.

I am minded to accept this as is, as I think there's nothing wrong with your code, it's a lower level issue with `juliac`.

graeme-a-stewart:
I am also wondering about how we test this - and that's a tricky question when OS X compilation is broken. Do we add a test that just early aborts with a false success on OS X?

m-fila:
> I am minded to accept this as is, as I think there's nothing wrong with your code, it's a lower level issue with `juliac`.

I don't have MacOS to check but perhaps this is originating from `LoopVectorization` - if I'm not mistaken we access `HostCPUFeatures` only through it.


> I am also wondering about how we test this - and that's a tricky question when OS X compilation is broken. Do we add a test that just early aborts with a false success on OS X?

Currently we have only test that the `@ccallable` functions work when called through Julia. It'd be good to  test compilation and then add C-executable checking that results are correct. I think on OS X, we could just mark it as `broken` and expect failure.

graeme-a-stewart:
> > I am minded to accept this as is, as I think there's nothing wrong with your code, it's a lower level issue with `juliac`.
> 
> I don't have MacOS to check but perhaps this is originating from `LoopVectorization` - if I'm not mistaken we access `HostCPUFeatures` only through it.

Would be worth checking that - perhaps @HarshitNagpal29 could take a look here?

> 
> > I am also wondering about how we test this - and that's a tricky question when OS X compilation is broken. Do we add a test that just early aborts with a false success on OS X?
> 
> Currently we have only test that the `@ccallable` functions work when called through Julia. It'd be good to test compilation and then add C-executable checking that results are correct. I think on OS X, we could just mark it as `broken` and expect failure.

So, should we accept this PR as is, as a baseline and then create a new issue for the improvements, or would you like to keep this open and add the tests?

HarshitNagpal29:
> > > I am minded to accept this as is, as I think there's nothing wrong with your code, it's a lower level issue with `juliac`.
> > 
> > 
> > I don't have MacOS to check but perhaps this is originating from `LoopVectorization` - if I'm not mistaken we access `HostCPUFeatures` only through it.
> 
> Would be worth checking that - perhaps @HarshitNagpal29 could take a look here?

I checked this on macOS arm64 with Julia 1.12.6. The PR branch instantiates/precompiles, but the `juliac` build fails during compilation with the same LLVM vscale error from HostCPUFeatures/cpu_info_aarch64.jl.

m-fila:
Thanks for double checking @HarshitNagpal29. Could you also try if by any chance removing `LoopVectorization` and `@turbo`  from here https://github.com/JuliaHEP/JetReconstruction.jl/blob/24403f2fcd63883a3195f9201aa67558c5718c00/src/Utils.jl#L121 eliminates the problem?

HarshitNagpal29:
> Thanks for double checking @HarshitNagpal29. Could you also try if by any chance removing `LoopVectorization` and `@turbo` from here
> 
> https://github.com/JuliaHEP/JetReconstruction.jl/blob/24403f2fcd63883a3195f9201aa67558c5718c00/src/Utils.jl#L121
> 
> eliminates the problem?

Ok, I tried it: removed `@turbo`, removed `LoopVectorization`. After resolving the compile environment, the `juliac` build succeeds.

graeme-a-stewart:
> > Thanks for double checking @HarshitNagpal29. Could you also try if by any chance removing `LoopVectorization` and `@turbo` from here
> > https://github.com/JuliaHEP/JetReconstruction.jl/blob/24403f2fcd63883a3195f9201aa67558c5718c00/src/Utils.jl#L121
> > 
> > eliminates the problem?
> 
> Ok, I tried it: removed `@turbo`, removed `LoopVectorization`. After resolving the compile environment, the `juliac` build succeeds.

Great! Thanks for looking at this.

Given that we do not have active users of the compiled version of the code, and that it's unlikely to be of interest to precompile on OS X, we simply note this in the docs, rather than complicating the code with a workaround for juliac + OS X.

m-fila:
> Given that we do not have active users of the compiled version of the code, and that it's unlikely to be of interest to precompile on OS X, we simply note this in the docs, rather than complicating the code with a workaround for juliac + OS X.

yes yes, it was just for understanding where the problems is so we can report it later

graeme-a-stewart:
I made an update to the OS X docs, so this is now good to go

PR #243, opened 2026-05-18
List public enums under public API docs

m-fila:
`RecoStrategy`, `JetAlgorithm`,` RecombinationScheme`  enums are public and already have docstring but none of them was included under public API in documenter generated docs. Now they will be listed properly.


Moelf (approved):

graeme-a-stewart (approved):
Thanks @m-fila ! Yes, this is correct,

PR #244, opened 2026-06-04
added-stats-plot-ext

HarshitNagpal29:
Adds benchmark timing statistics helpers to JetReconstruction.jl and moves terminal plotting behind a UnicodePlots package extension.

This keeps the core package free of a hard plotting dependency while allowing users who load UnicodePlots to call `plot_trial_times` for benchmark timing diagnostics.


graeme-a-stewart:
Hello @HarshitNagpal29 - first comment is that you should always provide a description of your pull request! It's important to record this, so that people looking into the history of the code have an idea of what this is about (including our future selves).

graeme-a-stewart:
Oh, as a minor P.S., yes you can develop on your `main` branch, but I like to keep that pristine and actually create a proper branch named for the changes that were developed

Project.toml
```diff
@@ -39,10 +42,12 @@ LorentzVectorBase = "0.1.0"
 LorentzVectorHEP = "0.1.6"
 Makie = "0.20, 0.21, 0.22, 0.23, 0.24"
 MuladdMacro = "0.2.4"
+Printf = "1.11.0"
```
graeme-a-stewart: This is too harsh a restriction on the package version. The need to support Julia LTS (=1.10.0) means you need to relax this constraint. For our purposes, even `Printf = "1.0"` should be fine here.

Project.toml
```diff
@@ -39,10 +42,12 @@ LorentzVectorBase = "0.1.0"
...
 Random = "1.10"
 Statistics = "1.10"
 StructArrays = "0.6.18, 0.7"
 Test = "1.10"
+UnicodePlots = "3.8.3"
```
graeme-a-stewart: Also check compatibility here - we shouldn't need the latest greatest version...

ext/JetBenchmarkPlots.jl
```diff
@@ -0,0 +1,37 @@
+module JetBenchmarkPlots
+
+using JetReconstruction
+using UnicodePlots
+
+function JetReconstruction.plot_trial_times(trial_timing::AbstractVector{<:Real})
```
graeme-a-stewart, line 14: No documentation! Please ensure that all methods have documentation in the usual Julia style (see any of the other code files for examples).

ext/JetBenchmarkPlots.jl
```diff
@@ -0,0 +1,37 @@
...
+    stats = JetReconstruction.summarise_trial_times(trial_timing)
+
+    bin_width = ceil(2 * stats.iqr / stats.n_samples^(1 / 3))
```
graeme-a-stewart, line 19: This is probably worth a comment! Here we are using the Freedman-Diaconis Rule to calculate the bin width, which works robustly on data with outliers,

graeme-a-stewart (changes requested):

graeme-a-stewart:
@HarshitNagpal29 - thank you! This looking nice, but needs a polish:

- Documentation, for methods and in `docs/visualisation.md`
- Convert the examples to use these new factorised methods
- Relax the package constraints on `Printf` and `UnicodePlots` (preventing Julia LTS compatibility)
- Add an example, which would be a test of `instrumented-jetreco.jl` to `tests/test-basic-examples.jl` (actually we should have had one of these already, it seems we only are testing `jetreco.jl`)

HarshitNagpal29:
> Oh, as a minor P.S., yes you can develop on your `main` branch, but I like to keep that pristine and actually create a proper branch named for the changes that were developed

Yes sir, i know that and I always develop on different branch, but it was minor fix so didn'nt think of making new branch but will remeber from next time. 😅

test/test-basic-examples.jl
```diff
@@ -2,12 +2,20 @@
...
+    @test success(run(pipeline(`$JULIA_CMD --project=$EXAMPLES_PROJECT $(@__DIR__)/../examples/jetreco.jl --algorithm=EEKt -R 1.0 -p -1.0 $(@__DIR__)/../test/data/events.eeH.hepmc3.zst`,
                                devnull)))
-    @test success(run(pipeline(`julia --project=examples $(@__DIR__)/../examples/jetreco.jl --algorithm=Durham $(@__DIR__)/../test/data/events.eeH.hepmc3.zst --dump test-output-ee.csv`,
+    @test success(run(pipeline(`$JULIA_CMD --project=$EXAMPLES_PROJECT $(@__DIR__)/../examples/jetreco.jl --algorithm=Durham $(@__DIR__)/../test/data/events.eeH.hepmc3.zst --dump test-output-ee.csv`,
                                devnull)))
     rm("test-output-ee.csv"; force = true)
 end
+
+@testset "Basic instrumented-jetreco.jl reconstruction example" begin
+    @test success(run(pipeline(`$JULIA_CMD --project=$EXAMPLES_PROJECT $(@__DIR__)/../examples/instrumented-jetreco.jl --maxevents=1 --nsamples=1 --algorithm=CA -R 0.4 $(@__DIR__)/../test/data/events.pp13TeV.hepmc3.zst`,
```
graeme-a-stewart: You'll need to run this code with `--plot` to exercise the unicode plotting code and so you should also use more samples (like 16 would be enough)

graeme-a-stewart: That should attack the codecov weaknesses above, that the code isn't exercised.

graeme-a-stewart (changes requested):
Just need to get the codecov numbers looking better.

We aim that 100% of all new lines of code are tested on a PR.

HarshitNagpal29:
@graeme-a-stewart Sir thanks for your suggestions, CI was failing because the example scripts are run in the separate `examples` project using `--project=examples`, but that environment was never instantiated during tests. So packages listed in `examples/Project.toml`, such as ArgParse and UnicodePlots, were declared but not installed/resolved in CI before the examples were executed.

I added an explicit examples-environment instantiation step in `test/test-basic-examples.jl` before running the example scripts. Since the test now uses `Pkg.instantiate()`, I also added `Pkg` as a test extra/target in `Project.toml`.

The downgrade workflow also needed adjustment: `Pkg` and `Printf` are Julia stdlibs, so they should not be downgraded by `julia-downgrade-compat`.

graeme-a-stewart:
Hello @HarshitNagpal29 

We still seem to have a few niggling issues to look into - can you check? 

FYI, for formatting please see https://juliahep.github.io/JetReconstruction.jl/stable/contributing/#Formatting.

Thanks

Graeme

P.S. I have invited you as a developer to the repo, so that the CI will run automatically for your PRs.

graeme-a-stewart (approved):

graeme-a-stewart:
@allcontributors please add @HarshitNagpal29 for code doc

PR #248, opened 2026-06-12
Fix spurious changes to examples Project.toml

graeme-a-stewart:
The PR in #244 fixed the setup of the `examples` environment for testing, but had the unfortunate side effect of writing the dev path to JetReconstruction as an *absolute* path, which is both a spurious change and a broken one for any other machine.

This PR fixes that so that the "develop" path for JetReconstruction in examples is always a relative one, viz. `".."`.


HarshitNagpal29 (approved):
LGTM

PR #249, opened 2026-06-12
Update to EDM4hep v1.0

graeme-a-stewart:
Include support fo the 1.0 release of EDM4hep.jl.

Conversion to JetReco types is now supported for both MCParticle and ReconstructedParticle types (previously only ReconstructedParticle was supported).

In the example file SimpleRecoEDM4hep.jl take the input file from the EDM4hep package, so there is no longer a dangling dependence on a local file.

A tiny "dump" script, QuickLookEDM4hep.jl is added for debugging.


m-fila (approved):
I checked with some EDM4hep v1 dummy input files and it works fine.

For posterity, there was a [proposal](https://codimd.web.cern.ch/borHwXH4TrCGdJY84r0oEA#EDM4hepJet-type-proposal-by-EIC) from EIC to extend the EDM4hep data model with a dedicated jet type similar to the type they have in [edm4eic](https://github.com/eic/EDM4eic/blob/f8f9680654deab04f30214a9270627719356ba00/edm4eic.yaml#L576). But since nothing was added so far just having support for `ReconstructredParticle` and `MCParticle` seems fine

graeme-a-stewart:
> I checked with some EDM4hep v1 dummy input files and it works fine.

Thanks!

> 
> For posterity, there was a [proposal](https://codimd.web.cern.ch/borHwXH4TrCGdJY84r0oEA#EDM4hepJet-type-proposal-by-EIC) from EIC to extend the EDM4hep data model with a dedicated jet type similar to the type they have in [edm4eic](https://github.com/eic/EDM4eic/blob/f8f9680654deab04f30214a9270627719356ba00/edm4eic.yaml#L576). But since nothing was added so far just having support for `ReconstructredParticle` and `MCParticle` seems fine

Indeed, it would be easy enough to add support for any additional jet-like types - hopefully they would support the same pattern for field accessors.

PR #247, opened 2026-06-11
External type generalisation

graeme-a-stewart:
This PR adds support so that the underlying *numerical* type of jets becomes parameterised, instead of being fixed to `Float64`. The parameterisation is `T <: Real`, which includes all float types and dual numbers (for potential autodiff applications).

One exception to parametrisation is the `TiledJet` structure used in the `N2Tiled` algorithm. This remains fixed to `Float64`, due to the use of a concrete singleton (`noTiledJet`) as the sentinel for linked lists (and to indicate tiles with no jets). As discussed in #91, this is tricky to optimise (I have a branch with a proof of concept that `TiledJet` can be parameterised, but it is not performance optimised as it allocates too many `Ref{TiledJet}`s).

# Main Change Summary

## Performance Report

The main optimisation steps of ensuring type consistency in mathematical operations have been done:

- Internal numerical values in functions are typed as `T`
- Function signatures for internal operations are uniformly typed to `T` when appropriate, to ensure that mixed parameters are not passed by accident

There has been some observation of a slight performance regression on pp reconstruction, but latest measurements show this as minimal O(1%) and not very consistent. Degradations also seem to be platform and phase of the moon (time) dependent, indicating no serious issue with the implementation.

## Parametrisation

`EEJet`, `PseudoJet` and `EERecoJet` are now parameterised on `T <: Real` (also the common abstract `FourMomentum` type). The number of constructors is considerably expanded to cover cases when initialisation is *explicit* as well as *implicit*, and including mixed types, where promotion is applied. Conversion from `LorentzVector`, `LorentzVectorCyl` and anything supporting `LorentzVectorBase` are also covered. If the EDM4hep extension is loaded, construction from `EDM4hep.ReconstructedParticle` is supported.

`ClusterSequence`, `HistoryElement` and `SoftKiller` are likewise parameterised and will be initialised to the same `T` as the input particles.

For consistency `T` is used as the numerical type of the jet's parameters, and `J` is used for the jet type.

### Invalid Values

When an invalid jet is returned from substructure functions, this is no longer a singleton (which cannot be parameterised), but is a `PseudoJet{T}` where `isvalid(jet) == false`. Technically this is implemented by setting the `_cluster_hist_index` to `typemin(Int)`, but this is an internal detail - use `isvalid()`!

## Reconstruction Interfaces

[...]

PR #252, opened 2026-07-29
n2tiled-improved-scaling-on-multithread

HarshitNagpal29:
## Summary

This PR adds an opt-in reusable workspace for full N2Tiled reconstruction, aimed at improving multithreaded event-processing scalability by reducing per-event allocations and garbage-collection pressure.

## Changes

- Add `N2TiledWorkspace` for reusing jets, complete clustering history, tiling data, nearest-neighbour arrays, and inclusive-jet output.
- Add callback-scoped `with_n2tiled_reconstruction`.
- Add in-place `inclusive_jets!`.
- Add explicit workspace-capacity release.
- Preserve the existing owning `jet_reconstruct` and `tiled_jet_reconstruct` behaviour.
- Add correctness, lifetime, concurrency, capacity-reuse, and allocation-regression tests.
- Document workspace ownership and one-workspace-per-worker usage.

The workspace path still produces a complete `ClusterSequence`; users can perform normal history, constituent, exclusive-jet, and inclusive-jet queries inside the callback.

## Compatibility

This is an additive API change. Existing users continue receiving independently owned results. High-throughput applications can opt into reusable storage and copy only the values that must be retained.

## Validation

- Workspace and owning paths produce exactly equal clustering results across multiple algorithms, radii, event sizes, and recombination schemes.
- Warmed allocation was reduced from approximately 416 kB/event to 640 bytes/event in the local regression measurement.


m-fila:
In case you were using different version of formatter https://github.com/JuliaHEP/JetReconstruction.jl/blob/main/docs/src/contributing.md#formatting

m-fila:
I pushed to your branch to remove the unrelated formatting changes. Please pull to synchronize and avoid conflicts

HarshitNagpal29:
> I pushed to your branch to remove the unrelated formatting changes. Please pull to synchronize and avoid conflicts

Thank a lot sir.

src/TiledAlgoLL.jl
```diff
@@ -96,12 +96,20 @@ This function sets the eta, phi, kt2, jets_index, NN_dist, NN, tile_index, previ
...
-    @inbounds jet.eta = rapidity(clusterseq.jets[jets_index])
-    @inbounds jet.phi = phi(clusterseq.jets[jets_index])
-    @inbounds jet.kt2 = pt2(clusterseq.jets[jets_index]) > 1.e-300 ?
-                        pt2(clusterseq.jets[jets_index])^p : 1.e300
+function tiledjet_set_jetinfo!(jet::TiledJet,
+                               jets::AbstractVector{PseudoJet},
+                               tiling::Tiling,
+                               jets_index::Int,
+                               R2,
+                               p)
```
m-fila, line 104: Perhaps I'm missing something but what is the benefit of `tiled_set_jetinfo!` operating on vector of jets rather than cluster sequence? If I'm not mistaken in all the calls the vector of jets is taken directly from cluster sequence, like this
```julia
tiledjet_set_jetinfo!(jet, clusterseq.jets, tiling, ijet, R2, p)
```

src/TiledAlgoLL.jl
```diff
@@ -187,20 +218,37 @@ function set_nearest_neighbours!(clusterseq::ClusterSequence, tiling::Tiling,
         end
     end
 
-    # Now create the diJ (where J is i's NN) table - remember that
+    # Now create the dij (where J is i's NN) table - remember that
```
m-fila:
```suggestion
    # Now create the dij (where j is i's NN) table - remember that
```

src/TiledAlgoLL.jl
```diff
@@ -187,20 +218,37 @@ function set_nearest_neighbours!(clusterseq::ClusterSequence, tiling::Tiling,
...
+            jetA = tiledjets[i]
+
+            dij[i] = _tj_diJ(jetA)
+            NNs[i] = jetA
+            jetA.dij_posn = i
+        end
     end
-    NNs, diJ
+
+    return nothing
```
m-fila, line 220: Should this return nothing or neighbours and diJ? According to the docstring it returns values - should it be updated?

src/TiledAlgoLL.jl
```diff
@@ -539,7 +627,7 @@ function _tiled_jet_reconstruct!(particles::AbstractVector{PseudoJet};
                         if jetI != jetB
                             jetI.NN_dist = dist
                             jetI.NN = jetB
-                            dij[jetI.dij_posn] = _tj_diJ(jetI) # update diJ...
+                            dij[jetI.dij_posn] = _tj_diJ(jetI) # update dij...
```
m-fila: I'm not sure if attempt at renaming`diJ` to `dij` is necessary in this PR. It adds a bit of noise but is still partial as there are still inconsistency like here a method is still using the `diJ` version

src/TiledAlgoLLStructs.jl
```diff
@@ -226,6 +238,259 @@ const tile_right = 1
...
+
+A workspace must never be used concurrently or reentrantly.
+"""
+mutable struct N2TiledWorkspace{T}
+    scratch::TiledScratch
+    jets::Vector{PseudoJet}
+    history::Vector{HistoryElement}
+    inclusive_output::Vector{T}
+    in_use::Threads.Atomic{Int}
+end
```
m-fila:
```suggestion
end
```

The `in_use` field is redundant. It's only accessed`_acquire_n2tiled_workspace!` in `_acquire_n2tiled_workspace!` to throw exception if detected that multiple threads try to mutate the object. I don't think enforcing "no-concurrent-mutation" like this is a good idea. It should be user's responsibility to ensure they aren't mutating a shared state

src/TiledAlgoLLStructs.jl
```diff
@@ -226,6 +238,259 @@ const tile_right = 1
...
+borrowed. They are overwritten by the next reconstruction using the same
+workspace.
+
+A workspace must never be used concurrently or reentrantly.
+"""
+mutable struct N2TiledWorkspace{T}
+    scratch::TiledScratch
+    jets::Vector{PseudoJet}
+    history::Vector{HistoryElement}
+    inclusive_output::Vector{T}
```
m-fila: The name is unclear. Is it meant to be used only with `inclusive_jets!`?  Users should be allowed to do any other kind of post-processing on the results too.

Also how is this supposed to be used - from what I understood it's used as a reusable storage for results and isn't actually used by the algorithm. In that case it might make more sense to decouple and allow users to provide their own re-usable buffer for results if they wish

src/TiledAlgoUtils.jl
```diff
@@ -136,9 +141,12 @@ function determine_rapidity_extent(eta::Vector{T}) where {T <: AbstractFloat}
 
     minrap, maxrap
 end
+function determine_rapidity_extent(eta::AbstractVector{T}) where {T <: AbstractFloat}
+    determine_rapidity_extent(eta, length(eta))
+end
 
 """
-    setup_tiling(eta::Vector{T}, Rparam::AbstractFloat) where T <: AbstractFloat
+    setup_tiling(eta::AbstractVector{T}, Rparam::AbstractFloat) where T <: AbstractFloat
```
m-fila: `N` argument missing in the docstring

src/ClusterSequence.jl
```diff
@@ -19,6 +19,24 @@ const NonexistentParent = -2
 "Cluster recombined with beam"
 const BeamJet = -1
 
+# Julia 1.11 introduced the `shrink` keyword for `sizehint!`.
+#
+# On Julia 1.11 and later, explicitly disable shrinking.
+# On Julia 1.10, the existing Vector method already behaves grow-only.
```
m-fila: Is the part for Julia <1.11 correct? I thought it wasn't documented if `sizehint` only grows so in principle for some type it could be implemented with shrinking and be compliant

src/TiledAlgoUtils.jl
```diff
@@ -42,7 +42,7 @@ struct TilingDef
 end
 
 """
-    determine_rapidity_extent(eta::Vector{T}) where T <: AbstractFloat
+    determine_rapidity_extent(eta::AbstractVector{T}) where T <: AbstractFloat
```
m-fila: `N` arguement not in the docstring

src/TiledAlgoLLStructs.jl
```diff
@@ -226,6 +238,259 @@ const tile_right = 1
...
+
+The vectors grow when necessary and do not shrink for smaller later events.
+Only entries `1:N` belong to the current event.
+"""
+function ensure_capacity!(scratch::TiledScratch, N::Int)
+    N >= 0 || throw(ArgumentError("N must be non-negative, got $N"))
+
+    if length(scratch.eta) < N
+        resize!(scratch.eta, N)
+    end
```
m-fila: I think the `ensure_capacity` is a bit misleading since it operates on size (as in number of stored elements) rather than capacity (as in total length of underlying buffer).

src/TiledAlgoLL.jl
```diff
@@ -410,32 +462,67 @@ function _tiled_jet_reconstruct!(particles::AbstractVector{PseudoJet};
 
     # This will be used quite deep inside loops, so declare it here so that
     # memory (de)allocation gets done only once
-    tile_union = Vector{Int}(undef, 3 * _n_tile_neighbours)
+    if scratch === nothing
+        tile_union = Vector{Int}(undef, 3 * _n_tile_neighbours)
+        eta_local = Vector{Float64}(undef, N)
+        tiledjets = Vector{TiledJet}(undef, N)
+    else
+        ensure_capacity!(scratch, N)
```
m-fila: Would it make sense to have `ensure_capacity!` always `resize!` underlying vectors so they have desired size and there is no need to keep track of "actual size" `N` and rewrite other methods to accept `N` and use it to clip iteration over these vectors (`setup_tiling`, `set_nearest_neighbours!`)?

src/TiledAlgoLL.jl
```diff
@@ -410,32 +462,67 @@ function _tiled_jet_reconstruct!(particles::AbstractVector{PseudoJet};
...
-    for ijet in 1:length(particles)
-        _eta[ijet] = rapidity(particles[ijet])
+    for ijet in 1:N
+        @inbounds eta_local[ijet] = rapidity(particles[ijet])
     end
 
-    tiling = Tiling(setup_tiling(_eta, R))
+    setup = setup_tiling(eta_local, R, N)
+
+    tiling = if scratch === nothing
```
m-fila:
```suggestion
    tiling = if isnothing(scratch)
```

src/TiledAlgoLL.jl
```diff
@@ -410,32 +462,67 @@ function _tiled_jet_reconstruct!(particles::AbstractVector{PseudoJet};
 
     # This will be used quite deep inside loops, so declare it here so that
     # memory (de)allocation gets done only once
-    tile_union = Vector{Int}(undef, 3 * _n_tile_neighbours)
+    if scratch === nothing
```
m-fila:
```suggestion
    if isnothing(scratch)
```

src/TiledAlgoLL.jl
```diff
@@ -410,32 +462,67 @@ function _tiled_jet_reconstruct!(particles::AbstractVector{PseudoJet};
...
+        ensure_capacity!(scratch, N)
 
-    # Setup the initial history and get the total energy
-    history, Qtot = initial_history(particles)
+        tile_union = scratch.tile_union
+        eta_local = scratch.eta
+        tiledjets = scratch.tiledjets
+    end
+
+    history, Qtot = if history_buffer === nothing
```
m-fila:
```suggestion
    history, Qtot = if isnothing(history_buffer)
```

src/TiledAlgoLL.jl
```diff
@@ -410,32 +462,67 @@ function _tiled_jet_reconstruct!(particles::AbstractVector{PseudoJet};
...
                                  Qtot)
 
     # Tiled jets is a structure that has additional variables for tracking which tile a jet is in
-    tiledjets = similar(clusterseq.jets, TiledJet)
-    for ijet in eachindex(tiledjets)
-        tiledjets[ijet] = TiledJet(ijet)
-        tiledjet_set_jetinfo!(tiledjets[ijet], clusterseq, tiling, ijet, R2, p)
+    for ijet in 1:N
+        @inbounds begin
+            if scratch === nothing
```
m-fila:
```suggestion
            if isnothing(scratch)
```

src/TiledAlgoLLStructs.jl
```diff
@@ -226,6 +238,259 @@ const tile_right = 1
...
+                        Float64[],
+                        Dict{Tuple{Int, Int}, TilingArrays}())
+end
+
+"""
+Reset and fill reusable recombination-jet storage.
+
+This mirrors the preprocessing behaviour of `tiled_jet_reconstruct`.
+
+When `preprocess === nothing`:
```
m-fila:
```suggestion
When `isnothing(preprocess)`:
```

src/TiledAlgoLL.jl
```diff
@@ -410,32 +462,67 @@ function _tiled_jet_reconstruct!(particles::AbstractVector{PseudoJet};
...
+            # Restore its current event-local initial identity.
+            jet.id = ijet
+
+            tiledjet_set_jetinfo!(jet, clusterseq.jets, tiling, ijet, R2, p)
+        end
     end
 
     # Now initialise all of the nearest neighbour tiles
-    NNs, dij = set_nearest_neighbours!(clusterseq, tiling, tiledjets)
+    if scratch === nothing
```
m-fila:
```suggestion
    if isnothing(scratch)
```

src/TiledAlgoLL.jl
```diff
@@ -261,13 +309,13 @@ updated and new nearest-neighbours must be run
...
 - `jetB`: The second jet.
-- `oldB`: The old second jet.
+- `oldB_tile_index::Int`: The tile index of the old second jet.
 - `tiling`: The tiling information.
 
 # Returns
 The number of neighbouring tiles added to the `tile_union`.
 """
-function find_tile_neighbours!(tile_union, jetA, jetB, oldB, tiling)
+function find_tile_neighbours!(tile_union, jetA, jetB, oldB_tile_index::Int, tiling)
```
m-fila, line 287: Changes in this function appear to be unrelated to the "workspace" and are meant to be a minor optimization meant to avoid copying "old second jet" in `_tiled_jet_reconstruct!` (caller of this)

src/ClusterSequence.jl
```diff
@@ -216,29 +260,68 @@ function add_step_to_history!(clusterseq::ClusterSequence, parent1, parent2, jet
...
+        if T == U
+            push!(output, jet)
+        elseif T <: LorentzVectorCyl
+            push!(output, lorentzvector_cyl(jet))
+        elseif T <: LorentzVector
+            push!(output, lorentzvector(jet))
+        else
+            error("Unsupported return type $T for inclusive jets")
+        end
+    end
```
m-fila: Is this the same logic as previously in `inclusive_jets` (not mutating), if so then I'd move the code as is and avoid small unrelated changes

docs/src/strategy.md
```diff
@@ -17,3 +17,40 @@ a `strategy` option to switch to a different option.
 For ``e^+e^-`` algorithms particle densities are low, so the only
 implementation for these algorithms is effectively of the same type as
 `N2Plain`.
+
+## Reusing N2Tiled storage
+
+The ordinary `jet_reconstruct` and `tiled_jet_reconstruct` interfaces return an
+independently owned `ClusterSequence`. This remains the right interface when a
+result must be retained.
```
graeme-a-stewart: Maybe not even this is true - I think it would be more efficient to use the workspace and to do a final *copy* (e.g., for substructure studies you need the whole ClusterSequence).

"This is the simplest interface to use, but it does carry a performance penalty, which the workspace described below overcomes."

docs/src/strategy.md
```diff
@@ -17,3 +17,40 @@ a `strategy` option to switch to a different option.
...
+
+High-throughput applications that process many independent events can avoid
```
graeme-a-stewart: Events are always independent, so just "High-throughput applications that process many events can avoid..."

docs/src/strategy.md
```diff
@@ -17,3 +17,40 @@ a `strategy` option to switch to a different option.
...
+
+for event in events
+    with_n2tiled_reconstruction(
+        workspace,
+        event;
+        algorithm = JetAlgorithm.AntiKt,
+        R = 0.4,
+    ) do clusterseq
+        jets = inclusive_jets!(workspace, clusterseq; ptmin = 5.0)
+        # Consume `jets` here, or copy it if it must outlive the callback.
```
graeme-a-stewart: I think I would even add the explicit `clusterseq` copy here as well (commented is fine), just to show clearly how to do it.

docs/src/strategy.md
```diff
@@ -17,3 +17,40 @@ a `strategy` option to switch to a different option.
...
+    end
+end
+```
+
+The callback receives a complete `ClusterSequence`, so history, constituent,
+and exclusive-jet queries remain available. Its jets and history, and the
+vector returned by `inclusive_jets!`, borrow storage from the workspace. They
+are overwritten the next time that workspace is used. Copy values that must
```
graeme-a-stewart: "You **must copy** values that will outlive the callback or use the ordinary owning interface."

docs/src/strategy.md
```diff
@@ -17,3 +17,40 @@ a `strategy` option to switch to a different option.
...
+outlive the callback or use the ordinary owning interface.
+
+A workspace must not be shared concurrently or used reentrantly. Parallel
+applications should create one workspace for each concurrent worker and keep
+each workspace owned by that worker. `release_n2tiled_workspace_capacity!` can
+be used after an unusually large event or when a long-lived worker should
+release retained storage.
```
graeme-a-stewart, line 60: I would here add an example or a link to a parallel processing example - that's the main use case that we envisage.

src/ClusterSequence.jl
```diff
@@ -216,41 +262,29 @@ function add_step_to_history!(clusterseq::ClusterSequence, parent1, parent2, jet
...
+The returned vector is the same object supplied as `output`.
 
-# Description
-This function computes the inclusive jets from a given `ClusterSequence` object.
-It iterates over the clustering history and checks the transverse momentum of
-each parent jet. If the transverse momentum is greater than or equal to `ptmin`,
-the jet is added to the array of inclusive jets.
+The output must not alias `clusterseq.jets`.
+"""
+function inclusive_jets!(output::Vector{T},
```
graeme-a-stewart: I am very unconvinced by the idea that we provide a version of `inclusive_jets` that reuses an output buffer. Generally these functions are exactly what's used to select and then process the output of the reconstruction. In addition, there tend to be few selections and these vectors are cheap to construct and pass back to the user. I would drop this version of the "borrowing" interface.

src/TiledAlgoLL.jl
graeme-a-stewart: To be fair, that was also the original function signature, but I think your point is correct @m-fila.

src/TiledAlgoLL.jl
```diff
@@ -123,30 +132,38 @@ end
...
+- `diJ::Vector{Float64}`: Reusable distance storage with the same length as
+  `tiledjets`.
 
 # Returns
-- `NNs::Vector{TiledJet}`: The vector of nearest neighbor jets.
-- `diJ::Vector{Float64}`: The vector of diJ values.
+- `nothing`
 
 The function iterates over each tile in the `tiling` and sets the nearest
 neighbor information for each jet in the tile. It then looks for neighbor jets
```
graeme-a-stewart: Small point, but for consistency, we should use _neighbour_ and it's variants (UK spelling).

src/TiledAlgoLL.jl
graeme-a-stewart: I don't think we ever would use a return value here - it's a mutating function with this call signature.

src/TiledAlgoLL.jl
graeme-a-stewart: It's still a valid optimisation (we are making a full copy of `jetB`, but only every using the tile index).

But I agree, it would be cleaner to apply this optimisation in its own PR (probably the first one!).

src/TiledAlgoLL.jl
```diff
@@ -410,40 +445,75 @@ function _tiled_jet_reconstruct!(particles::AbstractVector{PseudoJet};
 
     # This will be used quite deep inside loops, so declare it here so that
     # memory (de)allocation gets done only once
-    tile_union = Vector{Int}(undef, 3 * _n_tile_neighbours)
+    if isnothing(scratch)
```
graeme-a-stewart: I wonder if it would be possible in the case where `scratch` was not passed, that we create a temporary one here first - then we can avoid any logic switches and different code paths.

src/TiledAlgoLL.jl
```diff
@@ -410,40 +445,75 @@ function _tiled_jet_reconstruct!(particles::AbstractVector{PseudoJet};
...
+        ensure_length!(scratch, N)
+
+        tile_union = scratch.tile_union
+        eta_local = scratch.eta
+        tiledjets = scratch.tiledjets
+    end
 
-    # Setup the initial history and get the total energy
-    history, Qtot = initial_history(particles)
+    history, Qtot = if isnothing(history_buffer)
```
graeme-a-stewart: Likewise, just always make a `history_buffer`?

src/TiledAlgoLLStructs.jl
```diff
@@ -226,6 +238,195 @@ const tile_right = 1
...
+This mirrors the preprocessing behaviour of `tiled_jet_reconstruct`.
+
+When `isnothing(preprocess)`:
+
+- PseudoJet inputs are copied directly;
+- other supported input types are converted to PseudoJet.
+
+Otherwise, the supplied preprocessing function is called for every particle.
+"""
+function _prepare_recombination_jets!(jets::Vector{PseudoJet},
```
graeme-a-stewart: This code is less useful than the `construct_reco_jets()` function that is on the #247 PR.

(Tomorrow we can discuss this, but I think it's imperative that we merge #247 first here).

graeme-a-stewart (changes requested):
@HarshitNagpal29 - excellent work in designing something we can work with here.

I feel we have some way to go towards an optimal implementation with as few code path divergences in the implementation as possible (can we create a temporary workspace if there isn't a real one, then everything looks the same internally).

It also becomes more and more important to merge #247, as this PR is being broken every time we do a significant update of the code (currently it is broken by the fallout of #242). Having parameterised types is a big deal.

We also have to think about the convenient today of `jet_reconstruct(particles; algorithm=...`, which allows the user not to have to bother about the details of which reconstruction algorithm and strategy will be chosen. But here we start to move back to very explicit alg/strategy specific workspaces.

If we stick with this interface, we need to give q lot more examples of how to use the `with_n2tiled_reconstruction` for the most common cases.

And adding a multi-threaded example script in `examples` is essential.

src/ClusterSequence.jl
```diff
@@ -216,29 +262,20 @@ function add_step_to_history!(clusterseq::ClusterSequence, parent1, parent2, jet
...
-# Description
-This function computes the inclusive jets from a given `ClusterSequence` object.
-It iterates over the clustering history and checks the transverse momentum of
-each parent jet. If the transverse momentum is greater than or equal to `ptmin`,
-the jet is added to the array of inclusive jets.
-
-Valid return types are `LorentzVector` `LorentzVectorCyl` or the jet type of the
-input `clusterseq` (`U` - either `PseudoJet` or `EEJet` depending which
-algorithm was used).
+The returned vector is independently owned.
```
m-fila, line 266: Since there are no changes to `inclusive_jets` I think no changes to this docstring are needed either

src/TiledAlgoLL.jl
m-fila: Sorry I wanted to ask if any changes here are necessary since this this basically the same behaviour with just cosmetic changes. I think we don't need to change it

src/TiledAlgoLL.jl
```diff
@@ -557,3 +615,83 @@ function _tiled_jet_reconstruct!(particles::AbstractVector{PseudoJet};
...
+The ClusterSequence, its jets vector, and its history vector must not be retained
+for use after this workspace processes another event.
+
+Copy any data that needs to outlive the callback.
+
+The same workspace must not be used concurrently or reentrantly.
+Use ordinary `jet_reconstruct` or `tiled_jet_reconstruct` when an independently
+owned ClusterSequence is required.
+"""
+function with_n2tiled_reconstruction(f::F,
```
m-fila, line 642: I think at this point the callback based interface isn't really buying as anything. Perhaps the public interface could be just  `_n2tiled_reconstruct_with_workspace!`. I think it could be even renamed  just to `tiled_jet_reconstruct!` and have workspace as first argument - that would look almost the same as non-mutating interface

m-fila: I changed my mind. Perhaps it's the cleanest way to express "borrowing"

src/ClusterSequence.jl
```diff
@@ -69,35 +85,65 @@ function HistoryElement(jetp_index)
...
+    # A complete hadron-collider clustering sequence contains:
+    #
+    #   N initial history entries
+    # + N recombination/finalisation entries
+    # = 2N total entries.
+    _sizehint_for_reuse!(history, 2 * N)
 
-    Qtot::Float64 = 0
+    # Establish exactly N active initial-history slots.
+    resize!(history, N)
```
m-fila: Why the hint is for double the `N`? I think only `N` is used. The size_hint guarantees only the capacity will be at least `N` it may decide to give us more on it's own.

Also, is the size hint needed at all if the next thing is calling resize which can efficiently increase capacity on its own

src/TiledAlgoLL.jl
```diff
@@ -187,19 +204,33 @@ function set_nearest_neighbours!(clusterseq::ClusterSequence, tiling::Tiling,
...
+    NNs = similar(clusterseq.jets, TiledJet)
+    diJ = similar(clusterseq.jets, Float64)
+
+    set_nearest_neighbours!(tiling,
+                            tiledjets,
+                            NNs,
+                            diJ)
+
     NNs, diJ
 end
```
m-fila: Codecov is complaining that this method isn't accessed by any test.
If I'm not mistaken it isn't used anywhere with these arguments.
Since it isn't part of the public interface I don't there is any need to keep it for compatibility with previous versions

docs/src/strategy.md
```diff
@@ -17,3 +17,48 @@ a `strategy` option to switch to a different option.
 For ``e^+e^-`` algorithms particle densities are low, so the only
 implementation for these algorithms is effectively of the same type as
 `N2Plain`.
+
```
graeme-a-stewart: I'm not keen on this documentation being hidden under algorithm strategy, as this only applies to pp reconstruction. It should really be in its own section, where information on N2plain and e+e- can be added later as we merge in those fixes.

I would suggest we add a new section "Multithreaded Reconstruction" (or "High-performance Reconstruction"?) where we gather these documentation parts. And the obvious point should be made that for single threaded running, the standard interfaces work well, with only a small overhead.

docs/src/strategy.md
```diff
@@ -17,3 +17,48 @@ a `strategy` option to switch to a different option.
...
+## Reusing N2Tiled storage
+
+The ordinary `jet_reconstruct` and `tiled_jet_reconstruct` interfaces return an
+independently owned `ClusterSequence`. This is the simplest interface to use,
+but it does carry a performance penalty, which the workspace described below
```
graeme-a-stewart: Add "which becomes more severe in multi-threaded applications".

docs/src/strategy.md
```diff
@@ -17,3 +17,48 @@ a `strategy` option to switch to a different option.
...
+overcomes.
+
+High-throughput applications that process many events can avoid most per-event
```
graeme-a-stewart: "High-throughput *multi-threaded*..." "events *in parallel*"

docs/src/strategy.md
```diff
@@ -17,3 +17,48 @@ a `strategy` option to switch to a different option.
...
+temporary allocations with an `N2TiledWorkspace`:
```
graeme-a-stewart: "temporary allocations, *which greatly improves performance,* ..."

examples/n2tiled-multithreaded.jl
```diff
@@ -0,0 +1,106 @@
...
+    options = parse_command_line(args)
+    events = read_final_state_particles(options[:file], PseudoJet;
+                                        maxevents = options[:maxevents])
+
+    selected_jets = threaded_n2tiled_reconstruction(events;
+                                                    R = options[:distance],
+                                                    ptmin = options[:ptmin])
+
+    println("Processed $(length(events)) events with $(Threads.nthreads()) Julia threads; " *
+            "selected $(sum(length, selected_jets)) jets.")
```
graeme-a-stewart: add "inclusive"

src/ClusterSequence.jl
```diff
@@ -19,6 +19,22 @@ const NonexistentParent = -2
...
 const BeamJet = -1
 
+# Julia 1.11 introduced the `shrink` keyword for `sizehint!`. Disable shrinking
+# when that keyword is available. Older Julia versions have no public no-shrink
+# sizehint API, so leave their capacity untouched and let `resize!`/`push!` grow
+# it as needed.
+@inline function _sizehint_for_reuse!(buffer::Vector,
+                                      requested::Integer)
+    requested >= 0 ||
+        throw(ArgumentError("requested capacity must be non-negative, got $requested"))
```
graeme-a-stewart: This should *never* happen (only we call `_sizehint_for_reuse!`), but if it did happen please make it clear that this is an **internal** error. It's not a user ArgumentError. In fact add that a bug report should be filed!

src/JetReconstruction.jl
```diff
@@ -76,7 +76,8 @@ export plain_jet_reconstruct
 include("TiledAlgoUtils.jl")
 # Algorithmic part, tiled reconstruction strategy with linked list jet objects
 include("TiledAlgoLL.jl")
-export tiled_jet_reconstruct
+export N2TiledWorkspace, release_n2tiled_workspace_capacity!, tiled_jet_reconstruct,
```
graeme-a-stewart: Should `release_n2tiled_workspace_capacity!` be a public function? It seems unlikely that a user would want to use this, they would just let the workspace go out of scope and be normally garbage collected?

If someone really wants it they can access it in the package namespace.

src/TiledAlgoLL.jl
```diff
@@ -123,30 +132,38 @@ end
...
 Note: The diJ values are calculated as the kt distance multiplied by R^2.
 """
-function set_nearest_neighbours!(clusterseq::ClusterSequence, tiling::Tiling,
-                                 tiledjets::Vector{TiledJet})
+function set_nearest_neighbours!(tiling::Tiling,
+                                 tiledjets::Vector{TiledJet},
+                                 NNs::Vector{TiledJet},
+                                 diJ::Vector{Float64})
+    length(NNs) == length(tiledjets) ||
+        throw(ArgumentError("NNs and tiledjets must have the same length"))
```
graeme-a-stewart: Same comment about this being an internal error, not a user ArgumentError

graeme-a-stewart (changes requested):
Hello @HarshitNagpal29 - great, this is really converging. I think the code is good to go. I only have some comments about how we structure the documentation and a few of the docstrings.

docs/src/strategy.md
graeme-a-stewart: Also, this is worth a small mention/link from README.md

src/ClusterSequence.jl
graeme-a-stewart: As this was overlooked, I just pushed a commit to revert this

graeme-a-stewart (approved):
🎉

graeme-a-stewart:
@HarshitNagpal29 - I tweaked a few last documentation points (no point in ping-ponging them between us again).

Merging now - thanks so much for this! 👏

PR #256, opened 2026-08-16
n2plain-improved-scaling-on-multithread

HarshitNagpal29:
## Summary

This PR adds an opt-in reusable workspace for full N2Plain reconstruction, covering both hadron-collider and electron-positron algorithms.

The workspace reduces repeated per-event allocations by reusing reconstruction scratch arrays, jet storage, and clustering history while preserving the complete `ClusterSequence` semantics.

## Changes

- Add `N2PlainWorkspace` with support for:
  - `PseudoJet` storage for pp algorithms;
  - `EEJet` storage for e⁺e⁻ algorithms.
- Add callback-scoped `with_n2plain_reconstruction`.
- Reuse:
  - reconstructed jet storage;
  - complete clustering history;
  - pp nearest-neighbour and distance arrays;
  - e⁺e⁻ `StructArray` reconstruction storage.
- Resize scratch arrays to the exact active event length.
- Support caller-owned reusable output buffers through:
  ```julia
  inclusive_jets!(output, clusterseq)

## Compatibility

This is an additive API change. Existing users continue receiving independently owned results. High-throughput applications can opt into reusable storage and copy only the values that must be retained.


test/runtests.jl
```diff
@@ -19,7 +19,9 @@ function main()
 
     # New test structure, factorised tests for pp and e+e-
     include("test-pp-reconstruction.jl")
-    include("test-n2tiled-kernel.jl")
+
+    include("test-n2plain-kernel.jl")
+
```
m-fila:
```suggestion
    include("test-n2tiled-kernel.jl")
    include("test-n2plain-kernel.jl")
```

This should be a new test not replace the old one, right?

docs/src/multithreading.md
```diff
@@ -1,9 +1,10 @@
 # Multithreaded Reconstruction
 
-The ordinary `jet_reconstruct` and `tiled_jet_reconstruct` interfaces return an
-independently owned `ClusterSequence`. They are the recommended interfaces for
-ordinary, single-threaded reconstruction: their allocation overhead is small,
-and the returned result can be retained freely.
+The ordinary `jet_reconstruct`, `plain_jet_reconstruct`, and
+`tiled_jet_reconstruct` interfaces return an independently owned
```
graeme-a-stewart, line 4: Mention also `ee_genkt_algorithm` as well (gosh, those names look rather inconsistent now!)

docs/src/multithreading.md
```diff
@@ -38,6 +39,31 @@ from the workspace *and are overwritten by the next reconstruction using that
 workspace*. Copy any data that must outlive the callback or use an ordinary
 owning interface instead.
 
+For pp events where the N2Plain strategy is appropriate, the same workspace
+pattern is available through `N2PlainWorkspace`:
+
+```julia
+workspace = N2PlainWorkspace(PseudoJet)
```
graeme-a-stewart, line 46: See comment below re. how to parameterise this on algorithm, not jet type

docs/src/multithreading.md
```diff
@@ -38,6 +39,31 @@ from the workspace *and are overwritten by the next reconstruction using that
...
+        algorithm = JetAlgorithm.AntiKt,
+        R = 0.4,
+    ) do clusterseq
+        jets = inclusive_jets(clusterseq; ptmin = 5.0)
+        # Consume `jets` here, or copy any values that must be retained.
+    end
+end
+```
+
+`N2PlainWorkspace(EEJet)` provides the same reusable storage for e⁺e⁻
```
graeme-a-stewart, line 61: See comment below re. how to parameterise this on algorithm, not jet type

src/PlainAlgo.jl
```diff
@@ -404,3 +439,43 @@ function _plain_jet_reconstruct!(particles::AbstractVector{PseudoJet};
...
+                                                                          PPPlainScratch},
+                                              particles::AbstractVector;
+                                              algorithm::JetAlgorithm.Algorithm,
+                                              p::Union{Real, Nothing} = nothing,
+                                              R = nothing,
+                                              recombine = addjets_escheme,
+                                              preprocess = preprocess_escheme,
+                                              γ::Union{Real, Nothing} = nothing,
+                                              β::Union{Real, Nothing} = nothing)
+    is_pp(algorithm) ||
```
graeme-a-stewart, line 459: How could this error occur? `_n2plain_reconstruct_with_workspace!` is an internal function, not to be called by users.

src/PlainAlgo.jl
```diff
@@ -404,3 +439,43 @@ function _plain_jet_reconstruct!(particles::AbstractVector{PseudoJet};
...
+        throw(ArgumentError("algorithm $algorithm requires an EEJet N2PlainWorkspace"))
+    isnothing(β) ||
+        throw(ArgumentError("β is only supported by the Valencia algorithm"))
```
graeme-a-stewart, line 462: So here we have a parameter that *had to be `nothing`*, so then why does it exist as a parameter at all?

src/PlainAlgoStructs.jl
```diff
@@ -0,0 +1,207 @@
...
+    resize!(scratch, N)
+    return nothing
+end
+
+"""
+    N2PlainWorkspace([::Type{J} = PseudoJet])
+
+Reusable full-semantics storage for one N2Plain reconstruction worker.
+
+`J` selects the reconstruction family: use `PseudoJet` for pp algorithms and
```
graeme-a-stewart, line 61: OK, this is wrong - we should not parameterise on the jet type - a user could be working with any old jet type from their EDM, or, e.g., anything that supports `LorentzVectorBase`. PseudoJet and EEJet are not mandatory, we transform as needed.

Much better is to parameterise on the *algorithm*, which the user does know. I will make another comment re. how to do this generically.

graeme-a-stewart (changes requested):

graeme-a-stewart:
Great, thanks @HarshitNagpal29! My main technical comment here is that we should use the algorithm to parameterise the workspace, not the jet type.

One thing that seems to be missing, we have `with_n2plain_reconstruction` but what about `with_ee_reconstruction`? This should be added to support the same APIs with all the algorithms.

I would also say we should:

- Should add a convenience function `get_workspace(algorithm, event)` that returns the correct workspace for the algorithm/event pair (using the set particle density used for `Strategy::Best` in `jet_reconstruction()`)
  - And allow for a workspace to be passed to `jet_reconstruct`
- Should be documented
- Add borrow/MT test to `examples/benchmark.sh` and to `test.sh`

However, perhaps those last points can be a separate issue/PR, to keep this one clean and focused.

Issue #250, opened 2026-06-12
Switch to Runic formatting


graeme-a-stewart:
JuliaFormatter is having a few issues and we also got stuck at v1, because of v1/2 incompatibility.

As discussed in https://github.com/JuliaHEP/DistributionsHEP.jl/issues/48, a switch to Runic would now be preferred, with nice arguments summarised by @szabo137:

---

Indeed, I can 😸 

- simplicity (no config, no sync with other projects)
- better CI/CD integration (there is a maintained gh-action; JuliaFormatter needs some custom code around it, see [here](https://github.com/QEDjl-project/QEDbase.jl/tree/v0.2.0/.formatting))
- speed (installation and running are way faster for larger projects)
- active maintenance
- better editor integration
- better integration with git-hooks (e.g. via pre-commit)

and finally, a very cool badge [![code style: runic](https://img.shields.io/badge/code_style-%E1%9A%B1%E1%9A%A2%E1%9A%BE%E1%9B%81%E1%9A%B2-black)](https://github.com/fredrikekre/Runic.jl)

---

We will target the upcoming v2.0.0 release for this.

graeme-a-stewart:
Just to note, when this is done the commit should be added to `.git-blame-ignore-revs`.

PR #259, opened 2026-09-05
Switch to Runic formatting  #250

gitpsii:
This pull request completely modernizes the repository's formatting framework as suggested by the maintainers:

- **CI/CD Update:** Replaced the legacy 'julia-format' action with 'fredrikekre/runic-action@v1'.
- **Configuration Cleanup:** Removed the obsolete '.JuliaFormatter.toml' file since Runic utilizes zero-configuration rules set in stone.
- **Codebase Format:** Fully formatted the codebase files locally using the fast 'Runic.jl' engine to establish a clean formatting baseline.
- **Git History Preservation:** Added a '.git-blame-ignore-revs' tracking file.


.github/workflows/RunicFormat.yml
```diff
@@ -0,0 +1,13 @@
+name: Runic formatting
+on:
+  pull_request:
+
+jobs:
+  runic:
+    name: Runic
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
```
m-fila:
```suggestion
      - uses: actions/checkout@v7
```

.git-blame-ignore-revs
```diff
@@ -0,0 +1,2 @@
+﻿# Migrate codebase to Runic formatting
```
m-fila: Please don't add .git-blame-ignore-revs in this PR. We'll squash the commits and add blame ignore with correct hash later

m-fila (changes requested):
`contributing.md` in documentation and `AGENTS.md` should be updated as well

docs/src/contributing.md
```diff
@@ -76,9 +76,22 @@ If it takes a while to develop or implement review changes, rebase against
...
+```bash
+julia -e 'using Pkg; Pkg.Apps.add("Runic")'
+runic --inplace .
+```
+
+**For Julia v1.11 and earlier:**
+```bash
+julia --project=@runic -e 'using Pkg; Pkg.add("Runic")'
+~/.julia/bin/runic --inplace .
+```
```
m-fila: This doesn't work. This installs runic as a package not as an app so it won't appear there. Besides 1.11 has no apps

docs/src/contributing.md
```diff
@@ -76,9 +76,22 @@ If it takes a while to develop or implement review changes, rebase against
...
-[`JuliaFormatter`](https://github.com/domluna/JuliaFormatter.jl). Currently we
-use *version 1* of the formatter so please use this too or the CI may complain.
+### Formatting
+
+The code in this repository is formatted with [`Runic.jl`](https://github.com/fredrikekre/Runic.jl). Before submitting your pull request, please format your changes locally by running the appropriate command from the root directory of the project:
+
+**For Julia v1.12 and later:**
+```bash
+julia -e 'using Pkg; Pkg.Apps.add("Runic")'
+runic --inplace .
```
m-fila, line 85: This assumes `${HOME}/.julia/bin` is in a path which by default isn't. Any Julia app installation gives a warning but perhaps it should be mentioned here as well

docs/src/contributing.md
m-fila: This still refers to JuliaFormat

m-fila (approved):
I've checked changes in the source files, indeed this is just reformatting and Runic didn't break anything. Thank you

PR #261, opened 2026-09-07
test with `pre` instead of `nightly` in CI

m-fila:
I don't think we are that much of a bleeding edge to be concerned with nightlies, while some of the (optional) dependencies we have are quite fragile (CxxWrap) - anyway hopefully they will be fixed for release candidates which also give us information whether our code is expected to work in a future or not


graeme-a-stewart (approved):
Yes, I think this is more reasonable for our package. Thanks!

